PR #3, opened 2015-11-09
Tests

higs4281:
Slowly adding tests where there were none
## Additions
- test fixture
- model and search_index tests
- a convenience function for inspecting db objects
## Review
- @willbarton


higs4281:
have to park tests before big merge with front-end pipeline

PR #8, opened 2015-11-18
fix setups after traumatic reboot

higs4281:
This should set the standalone version on its feet again after the layout reboot. 
## Changes
- config and script adjustments to match new site layout
## Review
- @ascott1 @marteki

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,1 @@
+export DJANGO_SETTINGS_MODULE='paying_for_college.config.settings.standalone'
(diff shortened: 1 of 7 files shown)


ascott1:
This worked beautifully. Thanks @higs4281!

:+1: to merge

marteki:
YAY! This works great. :+1: from me as well.

ascott1:
MERGE IT

PR #9, opened 2015-11-18
Readmefix

higs4281:
Adds README section on setting up solr and college search
## Review
- @marteki @niqjohnson @ascott1


niqjohnson:
Not quite working for me. When I tried rebuilding the index, I got lots of errors like `Failed to add documents to Solr: [Reason: ERROR: [doc=paying_for_college.school.475705] unknown field 'autocomplete']`. 

Also, solr gives me errors like `33137 [qtp24433162-14] ERROR org.apache.solr.core.SolrCore  – org.apache.solr.common.SolrException: ERROR: [doc=paying_for_college.school.100636] unknown field 'autocomplete'`. I get no results for any college searches.

Also, a couple of other notes:
- Looks like the readme formatting is a little weird starting at `### Prepping solr` (it got included in the code block)
- I ran `solr -V` before the first step to get the path needed there, though it ended up matching the example you had exactly (`~/homebrew/Cellar/solr/4.10.2/example/schema.xml`). Then I ran `solr stop -all` before proceeding.

higs4281:
the indexing issue is a schema problem. i think i have a fix.
thanks for the formatting catch.

PR #10, opened 2015-11-18
Standalone

higs4281:
Our nascent front-end build will break standalone app pages at first, so this adds a backup static stash to keep pages alive during development.

Eventually, all PFC front-end resources should be generated by `setup.sh`
## Additions
- new directory:  /paying_for_college/local_static/
- configuration setting for using it when app is a standalone
## Changes
- update the README with solr setups
## Deletions
- content of /paying_for_college/static/  
  Assets should be built with setup.sh
## Review
- @mistergone @ascott1 @niqjohnson


higs4281:
this is wrapped up in newmodel PR

PR #11, opened 2015-11-20
Adds URL handling script to just-a-demo.html

mistergone:
Heyo! I didn't get to adding stuff to the template yet, but here are the important points of this PR (in the Additions section)!:
## Additions
- A new script! `query-handler.js` does work handling any query string in the url. Here's the skinny:
  - Inside `query-handler.js`, you will find an Array called `numericKeys`. The script will turn key found in that array into a number. So, in the query string `?tf=$1,000`, the value for `tf` will be turned into `1000`.
  - Inside `query-handler.js`, you will find an Object called `keyMaps`. So, if we want shorter urls, then we're abbreviate some things. Maybe `tuitionFees` will be abbreviated as `tf`. The script looks inside this object to know how to translate abbreviations into their longer forms. **Most importantly**, if a key doesn't appear in the `keyMaps` Object, it will not be returned by the script. So, in the query string `?tf=100&bk=99&ff=101` the key `ff` doesn't mean anything in `keyMaps`, so it won't be present in the Object the script returns.
- Included in `index.js` is an event handler that loads the URL query string values into the model,   performs the calculations, and updates the view.
- I also added an `init()` function. Right now, it calls `financials.calc()` to initialize that Object with the basic keys that the student debt calculator uses.
## Review
- @ascott1 @higs4281 @marteki @niqjohnson, check this out! Just pull down, open `just-a-demo.html`, and add a query string to the end, like... `?tf=100&rb=101&bk=99&tr=98&oe=102`

### DIFF
--- a/just-a-demo.html
+++ b/just-a-demo.html
@@ -21,7 +21,7 @@
     <input type="text" name="costs__books" data-financial="books">
   </label>
   <label for="costs__other">Other
-    <input type="text" name="costs__other" data-financial="other">
+    <input type="text" name="costs__other" data-financial="otherExpenses">
   </label>
   <label for="costs__transportation">Transportation
     <input type="text" name="costs__transportation" data-financial="transportation">
--- a/package.json
+++ b/package.json
@@ -45,7 +45,12 @@
     "vinyl-source-stream": "^1.1.0"
   },
   "dependencies": {
+    "chai": "^3.4.1",
     "format-usd": "^0.2.0",
+    "mocha": "^2.3.4",
     "student-debt-calc": "0.0.4"
+  },
+  "scripts": {
+    "test": "mocha test/js/"
   }
 }
--- a/src/disclosures/js/index.js
+++ b/src/disclosures/js/index.js
@@ -4,7 +4,13 @@ var $ = require( 'jquery' ),
     FinancialModel = require( './lib/financial-data.js' ),
     financials = new FinancialModel(),
     View = require( './lib/school-view.js' ),
-    schoolView = new View();    
+    schoolView = new View(),
+    queryHandler = require( './lib/query-handler.js' );
+
+function init() {
+  // Set the financials.values property with defaults
+  financials.calc();
+}
 
 function randomFinancials() {
   $( '[data-financial]' ).each( function() {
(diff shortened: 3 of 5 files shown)


mistergone:
## Update
- Added a JS testing suite (mocha with chai)
- Added the `npm test` command to `package.json`
- Added a barren test file in `test/js/test.js`

niqjohnson:
Awesome! Couple of things I noticed:
- `instituionalLoan` in `keymaps` is missing a t, I think
- Parent loans and work study can't be set in the query string. Was that intentional?
- Testing with this query string puts me in a whole lot of debt: `?tf=2222&rb=2222&bk=2222&tr=2222&oe=2222&sc=222&pg=222&sv=222&sv=222&fm=222&pk=222&ss=222&su=222&il=222&ir=222&pl=222&pr=222`. Total debt is $595,105 and monthly payment is $10,957,505. I get the expected debt and payment numbers if I type the values in the query string in by hand. (I'm working in the standalone app if that makes a difference.)

mistergone:
@niqjohnson - Yeah, some typos were expected... I'll fix that one real quick.

Parent loans and work study were just forgotten. I don't expect this actually represents a final accounting of possible inputs, it was just a starting point. But I'll add them real quick, too. :grin: 

Finally... the numbers you put in there included `222` for institutional loan rate and private loan rate. Since those numbers are expressed as decimals (0.079 is 7.9%), your loans have an interest rate of 22,200%. I'm honestly surprised you will only owe $11 million, and I strongly recommend you don't sign up for those loans. :smile: I didn't run this through the calculator to verify, but I think those numbers are correct. Trying `?tf=2222&rb=2222&bk=2222&tr=2222&oe=2222&sc=222&pg=222&sv=222&sv=222&fm=222&pk=222&ss=222&su=222&il=222&ir=.0222&pl=222&pr=.0222` gives a much more reasonable amount of debt.

Commit with the first two fixes coming soon!

mistergone:
Just as a note, we might want the urls to accept loan rates as "percents" meaning the url would contain `7.9` instead of `.079` for "7.9%". Things like this are easy to do, and might help with making the URLs easier.

niqjohnson:
@mistergone: Ah, decimal percentages, of course! Thanks for clearing that up.

The other fixes are working great for me. :+1:

higs4281:
:+1: sweet! can't wait to get this running in a template.

ascott1:
Excellent! Let's get this merged so that we can move it into a template.

Since @mistergone is gone on leave, we can address any other needed changes.

PR #12, opened 2015-11-20
Newmodel

higs4281:
Adds a model for handling disclosure text specified in settlements. 
Also steps back to using Django 1.6 and South
## Additions
- Disclosure text model and test
- UnityBox-friendly South migrations
## Review
- @marteki @niqjohnson @ascott1
## Testing
- if running standalone, pull in the newmodel branch, make sure you're in your virtualenv, then delete the sqlite database, uninstall django and re-run setup.

```
workon [YOUR ENV NAME]
rm paying_for_college/db.sqlite3
pip uninstall django  (answer yes to deleting files)
./local_setup.sh
```

give it a minute to load  
if it ends with "Installed 62323 object(s) from 1 fixture(s)," welcome to Django16landia.
- If you're in UnityBox, you're already in Django16landia, so you just need to run the migration. first use 
  `vagrant ssh` from the UnityBox directory to jump into the virtual box, then: 

``` bash
workon cfpb_django
cd /var/www/django/cfpb/
./manage.py migrate paying_for_college
```


niqjohnson:
Worked for me in both standalone and UnityBox. :+1:

PR #13, opened 2015-11-21
Urls

higs4281:
A simplified url scheme
## Removals
- debt/ and guides/ directories; we're all one happy PFC family now 
## Changes
- follows new url naming scheme
- fixes view test
- updates urls in nav_wrapper template
## Review
- @ascott1 @niqjohnson @marteki @kurzn 

You can ignore all but the last 2 commits


paying_for_college/config/settings/base.py
```diff
@@ -43,10 +44,10 @@
     'django.middleware.common.CommonMiddleware',
     'django.middleware.csrf.CsrfViewMiddleware',
     'django.contrib.auth.middleware.AuthenticationMiddleware',
-    'django.contrib.auth.middleware.SessionAuthenticationMiddleware',
+    # 'django.contrib.auth.middleware.SessionAuthenticationMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
-    'django.middleware.security.SecurityMiddleware',
+    # 'django.middleware.security.SecurityMiddleware',
```
ascott1, line 50: Do we want this commented out for now?

ascott1:
Small comment above, but overall :+1: from me. This puts us closer to getting the tool into the templates.

paying_for_college/config/settings/base.py
higs4281, line 50: we have to remove it, because django 1.6 can't use it and bombs.  
i leave it commented because I hope to return to 1.8 if a path to production opens up.  
also, these settings only affect the stand-alone version and are ignored elsewhere.

ascott1, line 50: Makes sense to me. Thanks!

niqjohnson:
`/paying-for-college/understanding-financial-aid-offers/` is 404 for me in the standalone version but working fine in UnityBox. Otherwise, looking good.

higs4281:
@niqjohnson a most excellent catch! fixed for standalone.

PR #14, opened 2015-11-24
Add responsive header and footer to standalone Django template

ascott1:
This adds a responsive version of the nemo header and footer (ala Owning a Home) when running the app in standalone mode.

Feel free to :-1: if we don't want to merge the demo template into `master`.
## Screenshot

screen shot 2015-11-24 at 12 39 11 pm
## Review
- @niqjohnson 
- @higs4281

### DIFF
--- a/paying_for_college/config/urls.py
+++ b/paying_for_college/config/urls.py
@@ -24,6 +24,9 @@ urlpatterns = [
     url(r'^managing-college-money/$',
         StandAloneView.as_view(template_name='manage_your_money.html'),
         name='pfc-manage'),
+    url(r'^demo/$',
+        StandAloneView.as_view(template_name='just-a-demo.html'),
+        name='pfc-demo'),
 ]
 
 if STANDALONE:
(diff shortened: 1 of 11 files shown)


higs4281:
:+1: works great in standalone. i vote to merge. we can adjust unitybox to match.

niqjohnson:
Looking awesome! In `paying_for_college/templates/just-a-demo.html`, can we override the viewport meta tag like so?

``` python
{% block page_viewport %}
  <meta name="viewport" content="width=device-width, initial-scale=1">
{% endblock %}
```

That was the only way I could get the responsiveness to happen in Chrome's device emulator (it worked fine shrinking down the window's width, though).

ascott1:
Good catch @niqjohnson! I updated the template and pushed it. Thanks!

niqjohnson:
:+1: Works for me in Chrome and Chrome's device emulator.

ascott1:
:tada:

PR #17, opened 2015-11-27
add view tests, turn off unneeded view code

higs4281:
Adding tests to the college cost-comparison tool.  
I'm looking for some python eyes outside my team, so pls forgive the user-tag spam.
## Additions
-  Tests, some new url configs and view tweaks.
## Removals
- Unneeded view code. 
## Review

py mates: take a peek if you can.
- @amymok @grapesmoker @imuchnik @rosskarchner @willbarton @fna

### DIFF
--- a/paying_for_college/disclosures/urls.py
+++ b/paying_for_college/disclosures/urls.py
@@ -10,10 +10,13 @@ urlpatterns = [
         name='pfc-feedback'),
 
     url(r'^technote/$',
-        TemplateView.as_view(template_name="technote.html"),
+        BaseTemplateView.as_view(template_name="technote.html"),
         name='pfc-technote'),
 
-    url(r'^api/search-schools.json', school_search_api),
+    url(r'^api/search-schools.json',
+        school_search_api,
+        name='school_search'),
+
     url(r'^api/bah-lookup.json', bah_lookup_api),
 
     url(r'^api/school/(\d+).json',
@@ -22,7 +25,7 @@ urlpatterns = [
 
     url(r'^api/worksheet/([1-z0-9-]*).json$',
         DataStorageView.as_view(),
-        name='worksheet'),
+        name='api-worksheet'),
 
     url(r'^api/worksheet/$',
         CreateWorksheetView.as_view(),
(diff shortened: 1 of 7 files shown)


paying_for_college/tests/test_views.py
```diff
@@ -1,29 +1,150 @@
...
+
+# http://www.consumerfinance.gov/paying-for-college/compare-financial-aid-and-college-cost/api/email/
+class EmailTest(django.test.TestCase):
+
+    fixtures = ['test_fixture.json']
+    factory = RequestFactory()
+
+    url = reverse('disclosures:email')
+    post_data = {'id': '00470019-e077-4fc3-9dbb-4a595fe976e6',
+                 'email': 'xxxfakexxx@gmail.com'}
```
marteki: We might not want to use three Xs here (just for optics and politics and all that fun stuff).

paying_for_college/disclosures/urls.py
```diff
@@ -10,10 +10,13 @@
...
     url(r'^technote/$',
-        TemplateView.as_view(template_name="technote.html"),
+        BaseTemplateView.as_view(template_name="technote.html"),
         name='pfc-technote'),
 
-    url(r'^api/search-schools.json', school_search_api),
+    url(r'^api/search-schools.json',
+        school_search_api,
+        name='school_search'),
+
```
willbarton, line 19: I realize this isn't really a change, but I'm going to comment on it anyway: is there a reason the endpoint ends with `.json`?

higs4281:
@willbarton we could ask ross; the url returns json, but `.json` isn't a flag for returning json as opposed to some other format. i suppose it sets up implementing `.csv` or `.xml` down the road, but it's an internal api, so that seems an unlikely use.

willbarton:
@higs4281 Wouldn't a query parameter make more sense for that then? `api/search-school?format=json`? I dunno, it's nit-picky, but it's something that sticks out to me.

The other thing that sticks out to me is the `POST` methods on the views. If we're going to taking in POST data, we'll need some kind of authentication/authorization mechanism. Would it make sense to be using something like Django Rest Framework here, which gives you all of that out of the box?

None of these are particular concerns for this PR, just things to consider going forward. This PR looks good. :+1:

higs4281:
@willbarton i think it would be worthwhile to refactor the api bits into restframework. thanks for the review.

PR #20, opened 2015-12-03
Add review offer section

niqjohnson:
Add "review your first year offer" section
## Additions
- Review section HTML
- Review section CSS/LESS
- Demo JS to populate review section fields
## Changes
- `.unit()` instead of straight ems in LESS
## Testing
- This will only work at small screen sizes in the standalone version. To test, pull in the `review-offer` branch, run `gulp`, then run `./manage.py runserver` from your virtual environment. The page will be at `http://localhost:8000/paying-for-college/demo/`. Fire up Chrome's device emulator (or shrink your browser) and check it out on a phone-sized screen.
- I've tested this in Mac Chrome and Mac Safari.
## Review
- @ascott1: This section is significantly more complex than the previous one. Does all the HTML and LESS look OK (especially the BEM classes?)
- @marteki
## Screenshots

verify-offer
## Todos
- Remove the demo JS when the real JS is working
- I know the HTML doesn't validate perfectly. That's because of the non-standard `autocorrect="off"` attribute on all the inputs. That attribute turns off the QuickType bar in iOS, which gives us a little more screen space when the keyboard is on the screen. Seems like a good tradeoff for non-validation to me, but I can remove that attribute if needed.
## Checklist


higs4281:
:+1: tested in standalone and in UnityBox. 
Tone boxes looked funky in Chrome's iphone6 emulator in standalone, but when I piped your branch through UnityBox and looked at it on a real iphone6, it looked fantastic. also good on an old ipad2.

UnityBox version's header still blows out, but that's not relevant to this PR.

src/disclosures/css/disclosures.less
```diff
@@ -50,6 +69,191 @@
...
+.line-item_title {
+  .grid_column(7);
+}
+
+.line-item_value {
+  .grid_column(5);
+}
+
+.line-item_amount {
+  float: right;
```
ascott1, line 196: This is nitpicky, but could we use the grid to align these rather than floating them?

ascott1:
Overall, this looks great. :clap:  @niqjohnson!

I have a small question above, but that doesn't even need to hold this up. :+1:

src/disclosures/css/disclosures.less
niqjohnson, line 196: Good question. I tried this a few ways, and this was the only way I could make all these constrains happen:
- Label is always flush left
- Number is always flush right
- Dollar sign sits 5 digits plus one comma away from the right edge
- Plus or minus sign sits just to the left of the dollar sign

The trouble I kept running into was the the dollar, plus, and minus signs don't quite fit to the grid. I have a feeling this might need to be adjusted to work in larger screen sizes anyway, so I'll keep noodling on it.

ascott1, line 196: That makes sense to me. I say onward and we can always revisit this as large screen designs make it our way.

marteki:
Looks great in standalone! Trying to get Unitybox working again so I can check on my Android phone.

marteki:
And working on my Android, thanks to Unitybox! :dancer:

PR #33, opened 2015-12-15
Introduce interface JS

ascott1:
This (begins to) connect the student debt calculator to the interface and adds an ESLint file for linting and some unit tests.
## Screenshot

screen shot 2015-12-15 at 5 16 41 pm
## Review

@marteki @niqjohnson @higs4281


marteki:
When running `gulp` in standalone, I get the following error:

> [09:13:49] Starting 'build'...
> [09:13:49] Finished 'build' after 8.19 μs
> [09:13:49] Starting 'default'...
> [09:13:49] Finished 'default' after 8.61 μs
> events.js:141
>       throw er; // Unhandled 'error' event
>       ^
> 
> Error: Cannot find module 'debounce' from '/Users/reedm/Repositories/cfpb/college-costs/src/disclosures/js/views'
>     at /Users/reedm/Repositories/cfpb/college-costs/node_modules/resolve/lib/async.js:46:17
>     at process (/Users/reedm/Repositories/cfpb/college-costs/node_modules/resolve/lib/async.js:173:43)
>     at ondir (/Users/reedm/Repositories/cfpb/college-costs/node_modules/resolve/lib/async.js:188:17)
>     at load (/Users/reedm/Repositories/cfpb/college-costs/node_modules/resolve/lib/async.js:69:43)
>     at onex (/Users/reedm/Repositories/cfpb/college-costs/node_modules/resolve/lib/async.js:92:31)
>     at /Users/reedm/Repositories/cfpb/college-costs/node_modules/resolve/lib/async.js:22:47
>     at FSReqWrap.oncomplete (fs.js:82:15)

Amusingly, the second Google result returned when I search for "Error: Cannot find module 'debounce'" has some familiar faces: https://github.com/cfpb/owning-a-home/issues/305 :stuck_out_tongue_winking_eye: Looks like an edit to package.json will fix it right quick.

ascott1:
Whoops! This should be fixed now.

niqjohnson:
This looks like a great start. I found a few weird things in my testing, but it seems like they might be better addressed in future PRs:
- I think `keyupListener()` will need some refinement to get rid of some unexpected behavior when tabbing into a field and when changing values in a field. Smoothing out that interaction definitely seems like a future PR to me, though.
- I'm getting some weird total debt numbers. For example, if I take out a $1000 Perkins and $1000 subsidized loan, the "Total federal loans" line item at the end of the federal loan section correctly shows $2000, but the "Total debt," "Your debt," and "Loans for first year" line items further down the page show $8042.92. I think those line items should also be $2000, and the $8042 is supposed to be the "Total cost of repayment with interest" value, maybe? Not sure if that's better addressed now or later.

I say :+1: to this PR as a foundation we'll start building on. So nice!

ascott1:
> I think keyupListener() will need some refinement to get rid of some unexpected behavior when tabbing into a field and when changing values in a field. Smoothing out that interaction definitely seems like a future PR to me, though.

Totally agree. I think a slight delay may help with this, but I can certainly do some tweaking.

> I'm getting some weird total debt numbers. For example, if I take out a $1000 Perkins and $1000 subsidized loan, the "Total federal loans" line item at the end of the federal loan section correctly shows $2000, but the "Total debt," "Your debt," and "Loans for first year" line items further down the page show $8042.92. I think those line items should also be $2000, and the $8042 is supposed to be the "Total cost of repayment with interest" value, maybe? Not sure if that's better addressed now or later.

I'll admit to not making it that far down the page yet :stuck_out_tongue: . I think it may be as simple as correcting the `data` attribute value.

niqjohnson:
> I'll admit to not making it that far down the page yet

In that case, a definite :+1: and we'll keep moving down the page.

ascott1:
I just pushed an update that will hook up all of the remaining fields, with the exception of anything that requires a monthly payment + interest calculation.

marteki:
Assuming remaining cost shouldn't be showing a value yet as well, then it's good on both standalone and Unitybox. :+1:

ascott1:
Remaining cost will be in included next PR.

PR #35, opened 2015-12-16
Api fixes

higs4281:
Updates api scripts and variables with Patrick's guidance
## Additions
- This is mostly adjusting variable notes as suggested by Patrick 
- Adds a try-accept block to capture output when there's an encoding problem with the data. 
  This function isn't used in production; it just lets me output custom CSV files for Patrick.
## Removals
- RETENTRATELT4 and RETENTRATE are removed at Patrick's direction.
## Review
- @ascott1


ascott1:
:+1: looks good to me!

PR #38, opened 2015-12-17
Model updates

higs4281:
Adjusts models and API script after consulting with Patrick
## Additions
- model fields to cover all API data points
## Changes
- adjust API scripts to match model changes
- @ascott1 @marteki


marteki:
Migrations were successful, and page loading fine in standalone and Unitybox both. :+1:

ascott1:
Pulled down locally and ran the migrations. Everything is working great. :+1:

PR #39, opened 2015-12-17
Accept URL parameters

ascott1:
This glues together the URL parameter work that @mistergone had done with the rest of the JS.
## Screenshot

screen shot 2015-12-17 at 3 54 29 pm
## Bonus (aka PR scope creep)
- Fixes the linting task and addresses all of the linting errors
- Calculates the remaining cost in UI and model
## Review

@niqjohnson & @marteki

### DIFF
--- a/src/disclosures/js/models/financial-model.js
+++ b/src/disclosures/js/models/financial-model.js
@@ -8,7 +8,7 @@ var financialModel = {
 
   init: function() {
     var val = getViewValues.inputs();
-    this.calc ( val );
+    this.calc( val );
   },
 
   calc: function( val ) {
@@ -25,14 +25,14 @@ var financialModel = {
 
     // total cost is attendance minus grants and scholarships
     model.totalCost =
-      model.costOfAttendance - model.totalGrantsScholarships
+      model.costOfAttendance - model.totalGrantsScholarships;
 
     // total contributions
     model.totalContributions =
       model.savings + model.family + model.workstudy;
 
     // total private loans
-    //TODO: accomodate for multiple private loans
+    // TODO accomodate for multiple private loans
     model.totalPrivateLoans =
       model.privateLoan + model.paymentPlan;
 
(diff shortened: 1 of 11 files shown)


ascott1:
It's worth noting that those parameters may need to change. I know @higs4281 has documented what they should be and those changes are easy to make.

src/disclosures/js/models/financial-model.js
```diff
@@ -7,8 +7,9 @@ var financialModel = {
   values: {},
 
   init: function() {
-    var val = getViewValues.inputs();
-    this.calc ( val );
+    var val = getViewValues.init();
+    console.log( val );
```
niqjohnson: Get rid of this `console.log`, or do we need to keep it in there for now?

niqjohnson:
This is looking good. One small line comment above and a couple of questions. I used these parameters:

```
?tf=20000&rb=1000&bk=500&tr=500&oe=500&sc=2000&pg=2000&sv=800&ws=800&fm=800&fl=3000&pk=3000&ss=1500&su=1500&il=500&ir=4&pl=1000&pr=6
```

Almost all the parameters populated the expected fields except for a couple of oddball ones:
- Should `pr` populate the private loan interest rate field?
- `il` and `ir` map to "loans from your school" and the associated interest rate in the current version of the tool. Should those map to "tuition payment plan from your school" and the associated interest rate in this version?

If we want to address those in a future PR, I say :+1: to this one.

higs4281:
:+1: to linting. i vote to rename vars after we get this on the boards.

src/disclosures/js/models/financial-model.js
ascott1: Oh, the shame.

ascott1:
@niqjohnson the changes I pushed should address your comments and we can rename variables as needed going forward.

niqjohnson:
:+1:

All the parameters in the URL are populating all the right fields for me. Awesome!

ascott1:
:tada:

PR #44, opened 2015-12-22
update repayment to reflect new income base repayment rules

ascott1:
This includes the updates to the repayment tool that were made in the old internal project repo.
## Review
- @higs4281 

I know things are set up differently between the two repos, so let me know if I mangled anything when transferring this over here.

### DIFF
--- a/paying_for_college/templates/repay_student_debt.html
+++ b/paying_for_college/templates/repay_student_debt.html
@@ -61,21 +61,7 @@
     </div>
   </div><!-- end .pfc-hero-wrapper -->
   <a class="cfpb_visually_hide focusable" href="#wizroot">Skip to main content</a>
-  <div class="pfc-nav-wrapper">
-    <nav>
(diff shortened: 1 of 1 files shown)


higs4281:
Content changes look great, but I think we'll want to stick with this static pathing:
  `{% static "paying_for_college/repaystudentdebt/ ..." %}`
That will be consistent with the rest of the app and newer apps on cfgov.
That pathing should make use of static assets restored in [PR47](https://github.com/cfpb/college-costs/pull/47)

ascott1:
Good catch @higs4281. That was unintentional when copying things between the old repo and here. I'll push a fix today.

ascott1:
@higs4281 I just pushed an update that should correct my static path error. Thanks again for catching!

higs4281:
:+1:

ascott1:
PUSH THE BUTTON

PR #49, opened 2015-12-29
Add verify info interactions with no JS

marteki:
Added no-JS friendly dynamic interaction for settlement students to verify their information, or signify that it isn't their information.
## Additions
- Added `div` wrappers for `info-right` and `info-wrong`
- Added `target` selectors to use when showing and hiding verification instructions
## Changes
- Changed HTML verification `buttons` to `a` elements, so the `target` attribute could be used.
- Moved instructions for what to do if a student receives the wrong offer information lower in the HTML code
## Testing
- Pull in the `no-js` branch and run `gulp`. Load it up, and try saying that it isn't your offer info, or that it is.
- Try it in standalone or in UnityBox. Try it on a phone or tablet.
- Try it in a browser with Javascript disabled, and see that it works there.
- Try it in IE8, to see that **it does not work there**. Luckily, the JS enhancement should, unless our IE8 user has also disabled JavaScript.

I tested it on Mac Chrome 47, Mac Safari 9.0, Mac Safari 9.0 with JS disabled, IE8, and Android Chrome Beta 48 on a phone.
## Review
- @niqjohnson @ascott1 @mistergone @higs4281 
## Todos
- Add JS to hide the button/link and some content
- Add keyframes animation for a smoother interaction
- Work with the other FEWD-like folks to see if we can restructure the page to have the Big Question buttons act the same way. I don't see a good way to do it currently without a more major restructuring of the page.
## Screenshots

On load:

screen shot 2015-12-29 at 4 39 22 pm
## Checklist

### DIFF
--- a/paying_for_college/templates/just-a-demo.html
+++ b/paying_for_college/templates/just-a-demo.html
@@ -86,95 +86,156 @@
 
     <section id="info-right" class="information-right">
         <div class="content_line"></div>
-    <div class="wrapper">
-        <section class="instructions step content_wrapper">
-            <div class="content_main">
-                <div class="instructions_wrapper">
-                    <div class="instructions_content instructions_content__right">
-                        <h2 class="step_heading">
-                            Before you are able to enroll, you will need to review the following sections:
-                        </h2>
-                        <ol class="list__flush-left">
-                            <li>
-                                Your financial aid offer
-                            </li>
-                            <li>
-                                Your offer evaluation
-                            </li>
-                            <li>
-                                Your options
-                            </li>
-                        </ol>
+        <div class="wrapper">
+            <section class="instructions step content_wrapper">
+                <div class="content_main">
+                    <div class="instructions_wrapper">
+                        <div class="instructions_content instructions_content__right">
+                            <h2 class="step_heading">
+                                Before you are able to enroll, you will need to review the following sections:
+                            </h2>
+                            <ol class="list__flush-left">
+                                <li>
+                                    Your financial aid offer
+                                </li>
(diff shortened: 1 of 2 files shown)


niqjohnson:
I like the direction this is going!

One thing that caught my eye: the changes in the `content-updates` branch are going to cause merge conflicts with the changes to `just-a-demo` here. We can figure out the best way to resolve those when everyone's back next week, though.

marteki:
@niqjohnson Yep, let's circle back on it early next week. My preference would be to get your changes in, then adjust this PR to not conflict with your code. But if there's a better way, we can figure it out on Monday or something.

ascott1:
> My preference would be to get your changes in, then adjust this PR to not conflict with your code

I think this is a good approach for this.

niqjohnson:
All the HTML updates are merged into `master` now. There were A LOT of updates, so it might be easier to scrap this PR and redo the HTML changes. Sorry! (Hopefully no more big HTML changes now, though.)

Also, with the changes, I think there's a chance we might have other `:target` elements on the page (like if we scroll the page up to step 1 from a link in step 3) that could be `display: inline-block` (i.e. they might be columns). So, we might want to change

``` css
*:target {
  display: block;
}
```

to something like

``` css
.progress-section__block:target {
  display: block;
}
```

marteki:
@niqjohnson I'll scrap this PR, but wait on changes to the worksheet template before touching the HTML/CSS/JS again.

Thanks for the heads-up about the `inline-block` targets! I'll adjust it to be more specific.

PR #57, opened 2016-01-12
add front-end build steps to local setup

higs4281:
Add front-end build to local_setup so README works
## Additions
- two lines to local-setup
## Review
- @mistergone @marteki


ascott1:
:+1:

PR #59, opened 2016-01-13
Run essential gulp tasks when starting watch

ascott1:
This was driving me nuts. Now running gulp watch will build all of the JS/CSS and copy necessary files on it's initial run. This means no more:

```
gulp
gulp watch
```

Now we should be able to just run `gulp watch` while we sit back with a :tropical_drink: 
## Review

@mistergone @niqjohnson @marteki


mistergone:
:+1: 

:heart: it

niqjohnson:
:+1:

Yeeeeeeeeeessssss!

PR #61, opened 2016-01-13
Map correct URL parameters

ascott1:
We were using abbreviated URL parameters from the demo and several were missing. This _should_ correct that and make sure they populate in the UI.
## Testing

Pull down the `school-api` branch (so named because this is a dependency of that work), run `gulp` and pass a URL such as:

`http://localhost:8000/paying-for-college2/understanding-financial-aid-offers/api/offer/?iped=408039
&pid=981&oid=f38283b5b7c939a058889f997949efa566c616c5&tuit=38976&hous=3000&book=650&tran=500&othr=500
&pelg=1500&schg=2000&stag=2000&othg=100&mta=3000&gib=3000&&wkst=3000&parl=10000
&perl=3000&subl=15000&unsl=2000&ppl=1000&gpl=1000&prvl=3000&prvi=4.55&insl=3000&insi=4.55`

Please try any combination of parameters with values from the documentation list [ghe pages/paying-for-college/dynamic-disclosures-documentation/#offer-data]
## Review

Any combination of the following lovely people:

@higs4281 @niqjohnson @mistergone @marteki


src/disclosures/js/utils/query-handler.js
```diff
@@ -11,28 +11,38 @@ function queryHandler( queryString ) {
...
+    perl: 'perkins',
+    subl: 'staffSubsidized',
+    unsl: 'staffUnsubsidized',
+    ppl:  'parentplus',
+    gpl:  'gradplus',
+    prvl: 'privateLoan',
+    prvi: 'privateLoanRate',
+    insl: 'institutionalLoan',
+    insi: 'institutionalLoanRate',
+    sav:  'savings'
```
higs4281: 'sav' is the only var I don't see in the url spec. Otherwise, looks perfect. :+1:

ascott1:
@higs4281 I removed `sav` in the last commit. Thanks for catching!

ascott1:
Updated to reflect corrections from Patrick:

> Here are a couple notes on this:
> - stag: 'state529plan' ---> 'stateGrants',
> - Remove the following-  
>   - fam
>   - sav

higs4281:
:+1:

niqjohnson:
Looking good. A few of things I noticed (they may be outside of the scope of this PR, though):
- I couldn't get the URL in the testing notes to work. My local instance doesn't have a route for either `/paying-for-college2/understanding-financial-aid-offers/api/offer/` or `/paying-for-college2/understanding-your-financial-aid-offer/api/offer/`. There's a very good chance this is my Django ineptitude showing. I checked everything in the UI instead, and most of the fields populated as expected.
- I think we might still need `sav`. I didn't see another way to populate the "Cash you will provide" field in the UI (`<input id="contrib__savings" name="contrib__savings" data-financial="savings">`)
- Is `gpl` meant to populate the "Direct PLUS loan" field (`<input id="contrib__direct-plus" name="contrib__direct-plus" data-financial="directPlus">`)? That wasn't working for me.
- I couldn't find a way to populate the private loan fees (`<input id="contrib__private-loan-fees" name="contrib__private-loan-fees" data-financial="privateLoanFees">`) and private loan grace period (`<input id="contrib__private-loan-grace-period" name="contrib__private-loan-grace-period" data-financial="privateLoanGracePeriod">`) fields.
- Private loan interest and payment plan interest show up as 0.079% for me using the query string in the testing notes.

ascott1:
@niqjohnson in response to your questions above:
1. I think you got the correct URL in the end. This should only impact the UI
2. Since savings isn't a parameter that the school will pass us it's been taken out of the URL and users can enter it themselves
3. This should be addressed now. Please ignore the lack of camelCase until the `student-debt-calculator` variable names are updated
4. These don't have documented URL parameters. They may either need to be entered or are part of the constants API.
5. Those are the default vaules pulled from the API: http://localhost:8000/paying-for-college2/understanding-your-financial-aid-offer/api/constants/

I hope that helps! Thanks for the thorough review!

niqjohnson:
@ascott1: Points 1-3 look good to me, thanks!

Point 4 (private loan fees and grace period) looks good for this PR, too. Whether or not those parameters should be included in the offer data is a separate issue, I think.

For point 5 (private loan and payment plan interest rates), I assumed the `prvi` and `insi` values were populating those numbers. I'm not totally up on what's happening with those fields, though, so I may well be misunderstanding.

So, unless there's work to be done on point 5, I say :+1: to merging.

marteki:
Loading from the URL is working (after adjusting to the right page stub and removing /api/offer/). I'm seeing weirdness in the summary fields, though, as shown in the following screenshots. This is post-reloading collegedata.json, in case some value was missing: 

screen shot 2016-01-14 at 2 59 12 pm

screen shot 2016-01-14 at 2 59 28 pm

screen shot 2016-01-14 at 2 59 42 pm

ascott1:
Hmm, that is a new bug introduced by the removal or re-naming of something in a recent commit. Fix coming.

ascott1:
`NaN` issues should now be addressed. Could a couple of folks pull this down and re-test?

niqjohnson:
The NaNs are gone for me, but the Direct PLUS input is back to not taking the value in the URL (might have to do with my line comment above?).

ascott1:
@niqjohnson is the URL parameter for the Direct PLUS loan `gradplus`? It should only accept a value if the program level is graduate. Right now we're assuming undergrad (I don't see a program level in the URL scheme though or an option to toggle it in the UI)

niqjohnson:
@ascott1: Ah, fancy! That makes sense. Everything looks good, then.

marteki:
NaNs are gone, and all values coming through. I see a bug around the totals of some of the groups, but it can be addressed in another PR.

:+1:

higs4281:
@ascott1 ok to merge this? would be great to get it on the build server for demos.

ascott1:
MERGING!

PR #62, opened 2016-01-14
Notifications

higs4281:
Adds notification functionality now that we've worked it out with the first school.
## Additions
- one view, one url, and two new models, plus tests
## Testing
- Can't be fully tested until js notification code is added
## Review
- @ascott1 @mistergone @vccabral
  migrations can be ignored
## Checklist

### DIFF
--- a/paying_for_college/models.py
+++ b/paying_for_college/models.py
@@ -284,6 +284,7 @@ class Notification(models.Model):
             'errors': self.errors
         }
         school = self.institution
+        # we prefer to use endpount notification, so use it first if existing
         if school.contact:
             if school.contact.endpoint:
                 requests.post(school.contact.endpoint, data=payload)
--- a/paying_for_college/models.py
+++ b/paying_for_college/models.py
@@ -5,6 +5,10 @@ except:  # pragma: no cover
     from ordereddict import OrderedDict
 import uuid
 import json
+from string import Template
+
+import requests
+from django.core.mail import send_mail
 
 HIGHEST_DEGREES = {  # highest-awarded values from Ed API
     '0': "Non-degree-granting",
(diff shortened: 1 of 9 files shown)


paying_for_college/views.py
```diff
@@ -273,3 +276,21 @@ def school_search_api(request):
...
 
     return HttpResponse(json_doc, content_type='application/json')
+
+
+class VerifyView(View):
+    def post(self, request):
+        if request.body:
+            timestamp = timezone.now()
+            data = json.loads(request.body)
+            school = School.objects.get(school_id=int(data['iped']))
```
vccabral: These line will give key errors that will not be caught, I believe.

paying_for_college/models.py
```diff
@@ -243,6 +266,41 @@ def primary_alias(self):
...
+            'errors': self.errors
+        }
+        school = self.institution
+        # we prefer to use endpount notification, so use it first if existing
+        if school.contact:
+            if school.contact.endpoint:
+                requests.post(school.contact.endpoint, data=payload)
+                return "school notified via endpoint"
+            elif school.contact.contact:
+                send_mail("CFPB disclosure notification",
```
vccabral, line 293: Send mail is awesome but I don't see any configs in settings.py. Is there a "college-costs" version of send_mail method? Are you going to send out more emails for any reason? If so, I would consider refactoring the send_mail into it's own module and use a custom send_mail method that gives you a custom handling. For example, you can create your custom mail_method to just accept a template name and a context and it can have a centralized template selection (or even django based templates). All emails will automatically come from no-reply@cfpb.gov etc.

vccabral:
@higs4281, by and large, a fantastic PR. I made some suggestions or offered areas I would think about but you have a lot more knowledge of the requirements of this specific application so LGTM in general and if you need any more PR reviews hit me up.

paying_for_college/models.py
higs4281, line 293: excellent idea; grazie. 
the current PFC tool is using send_mail; i'll check settings.

PR #63, opened 2016-01-14
Add basic handling of multiple private loans

mistergone:
This PR adds basic support for multiple private loans. As an important note, totals are not currently being calculated or displayed correctly. I decided that fix was out of scope for this PR, as it will involve another update to student-debt-calc. So, that will come soon!

**Known bug**: Removing every private loan makes it impossible to add new private loans. My question is, "Should the user be able to remove the last private loan"? When we answer this, I will implement a fix one way or the other.
## Additions
- Added user interface for adding and removing private loans
- Update version of student-debt-calc to support multiple private loans
- Updated dispatchers to handle multiple private loans
## Testing
- Create a branch locally called `chuckisgreat`, then switch to that branch and pull in my latest repo. Fire up the old `runserver` and then try to add and remove private loans!
## Review
- @ascott1 @marteki @niqjohnson @higs4281 
## Screenshots

screen shot 2016-01-14 at 11 11 15 am
## Todos
- Fix loan totals to display proper numbers
- Address "removing all private loans" bug, as detailed above
## Checklist
## Special Thanks

@ascott1 and @marteki and @niqjohnson  and @higs4281 :  **Thanks for doing all this great work in my absence!!!**
## Animated GIF

murdock-winks

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,6 @@
   },
   "dependencies": {
     "format-usd": "^0.2.0",
-    "student-debt-calc": "0.0.6"
+    "student-debt-calc": "0.1.2"
   }
 }
--- a/paying_for_college/templates/worksheet.html
+++ b/paying_for_college/templates/worksheet.html
@@ -172,7 +172,7 @@
                     <div class="content_line"></div>
                 </div>
             </section>
-            <section class="review step content_wrapper">
+            <section class="review step content_wrapper" data-section="review">
                 <div class="content_main">
                     <div class="review_wrapper">
                         <div class="evaluate_intro">
@@ -933,7 +933,7 @@
                                         </p>
                                     </div>
                                     <div class="private-loans">
-                                        <div class="private-loans_loan">
+                                        <div class="private-loans_loan" data-private-loan="0">
                                             <div class="private-loans_heading">
                                                 <div
                                                 class="private-loans_heading-text">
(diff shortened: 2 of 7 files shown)


ascott1:
@mistergone I'll add to the TODO that this doesn't update the loan totals in the UI, but otherwise works as expected.

ascott1:
Small comment above, otherwise :+1:

niqjohnson:
Styles are looking good at all the screen sizes I tested at.

I'm seeing a console error of `Uncaught TypeError: Cannot read property 'amount' of undefined` if I follow these (admittedly unlikely) steps, though:
1. Add a second private loan
2. Remove the **first** private loan
3. Change the value of any of the remaining private loan fields

Also, I think we decided that the user would be allowed to remove the last private loan. @caheberer, does that sound right to you?

marteki:
From our discussion on a private repo:

> - The "Add another private loan" button inserts another set of private loan fields (amount, rate, fees, grace period, term) below the last set.
> - **Will there be a "remove private loan" button on loans two, three, etc.?**
> 
> No JS: Display three sets of private loan sections by default.
> 
> Full JS: Display one by default, as well as the add private loan button. Also, add a remove private loan button.

~~I'm reading this as only loans two, three, or up would have a remove button.~~ Never mind, I was reading it wrong. For full JS, display one loan with both add and remove buttons.

higs4281:
:+1:  works as described

marteki:
Works as described. Seeing the same console error as Niq, but that's probably due to not all the JS being hooked up yet.

:+1: for merging from me, assuming that we work together to add functional test coverage with the JS hookups.

mistergone:
I'm going to merge this just to get it underway, then file issues about the following:
- Bug when adding then deleting a school, per @niqjohnson and @marteki 
- Create functional test coverage
- Enumerating IDs in each private loan
- NoJS view

PR #65, opened 2016-01-15
Adds an /offer/ url

higs4281:
Adds an /offer/ url that checks for errors and delivers school and program data to the page
## Additions
- new url, view, and tests
## Changes
- worksheet template tweaked to fill in basic school info and add schoolData and programData js objects to the DOM. 
## Testing
- after pulling in branch, run `./manage.py migrate paying_for_college` to get latest model changes
- for standalone, try hitting

```
http://localhost:8000/paying-for-college2/understanding-your-financial-aid-offer/offer/?iped=408039&pid=981&oid=f38283b5b7c939a058889f997949efa566c616c5&tuit=38976&hous=3000&book=650&tran=500&othr=500&pelg=1500&schg=2000&stag=2000&othg=100&ta=3000&mta=3000&gib=3000&wkst=3000&parl=10000&perl=3000&subl=15000&unsl=2000&ppl=1000&gpl=1000&prvl=3000&prvi=4.55&insl=3000&insi=4.55
```
- this PR also fixes the program api to accept schoolID-offerID pairing (hyphen between), so this url should get you program json:

```
http://localhost:8000/paying-for-college2/understanding-your-financial-aid-offer/api/program/408039-981/
```
## Review
- @niqjohnson @ascott1 @mistergone @marteki but in scrum order
## Screenshots

Should see something like this:

offer_sample
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 
 ## Unreleased
 - new url for handling querystring offer data
+- added notification functionality for disclosures
 
 ## 1.1.0 - 2015-10-28
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 All notable changes to this project will be documented in this file.
 We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 
+## Unreleased
+- new url for handling querystring offer data
 
 ## 1.1.0 - 2015-10-28
 
--- a/package.json
+++ b/package.json
@@ -52,6 +52,6 @@
   },
   "dependencies": {
     "format-usd": "^0.2.0",
-    "student-debt-calc": "0.0.6"
+    "student-debt-calc": "0.1.2"
   }
 }
--- a/paying_for_college/disclosures/urls.py
+++ b/paying_for_college/disclosures/urls.py
@@ -2,7 +2,12 @@ from django.conf.urls import url
 from paying_for_college.views import *
 
 urlpatterns = [
-    url(r'^$', BuildComparisonView.as_view(), name='worksheet'),
+    url(r'^$',
+        BuildComparisonView.as_view(), name='worksheet'),
+
+    url(r'^offer/$',
+        OfferView.as_view(), name='offer'),
+
     url(r'^api/email/$', EmailLink.as_view(), name='email'),
 
     url(r'^feedback/$',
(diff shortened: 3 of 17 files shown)


marteki:
Scrum order for today? Guess that means I'm up next.

Following testing instructions gets me the following offer verification info:

> Brown Mackie College-Fort Wayne
> Fort Wayne, IN
> SCHOOL TYPE
> For Profit
> PROGRAM
> Occupational Therapy Assistant - 981
> PROGRAM TYPE
> Bachelor's degree
> OFFER ID
> f38283b

We have a match!
:dancer: :+1: 

(I'm amused that the testing offer ID directly matches one of our commit IDs. Intentional?)

higs4281:
@marteki too observant!

niqjohnson:
Oh, nice! The school and program data in your screenshot are showing up for me in both standalone and UnityBox.

I do see a console error when the page loads, which I think is preventing any of the JS from working (not sure if it's related):

```
GET http://localhost:3000/paying-for-college2/understanding-your-financial-aid-offer/understanding-your-financial-aid-offer/api/constants/ 404 (Not Found)

get-api-values.js:17 something went wrong error Not Found
```

Otherwise, :+1:

(There are a few other places in the content where the school name is going to appear, but I think it makes sense to hook those instances up in a separate PR once we have revised content.)

higs4281:
@niqjohnson ah. this [relative-url api call](https://github.com/cfpb/college-costs/blob/master/src/disclosures/js/dispatchers/get-api-values.js#L8) works for /understanding-your-financial-aid-offer/ but i'm using the same template for /understanding-your-financial-aid-offer/offer/

would be trivial (on back end anyway) to avoid this api call and provide the constants on page render for `/offer/` requests.

higs4281:
@ascott1 any advice on handling this [relative-url api call](https://github.com/cfpb/college-costs/blob/master/src/disclosures/js/dispatchers/get-api-values.js#L8), which I'm breaking by using the same template for two urls. (see previous comment)

Could be a non-issue down the road, because we could use `/understanding-your-financial-aid-offer/offer/` url only for settlement disclosures and the bare `/understanding-your-financial-aid-offer/` for non-settlement,

ascott1:
@higs4281 I _think_ we could back it up to the root of the project and include the full path from `paying-for-college` forward.

higs4281:
@ascott1 i did that in final commit, but i also use the base_url setting value so when we drop the "2" from paying-for-college2 it will update automatically.

i mashed all the latest PRs together locally and it's glorious.

PR #82, opened 2016-01-26
Start making the content more dynamic

marteki:
Makes it so that values in the textual page content are dynamic for
- the school's name,
- the remaining cost, and
- the total debt including interest to be paid back.

Also changes the debt burden recommendation to ~~10%~~ 8% and removes a next step that's not longer going to be included in the tool.
## Additions
- Include dynamic values for school name, remaining cost, and debt plus interest in the text
## Removals
- Specific call-out for what values in the aid offer have been modified, and by how much
## Changes
- Debt burden recommendation moved to ~~10%~~ 8% (to align with Dept of Ed choices)
## Testing
- To test: pull in the `dynamic-content` branch and run gulp. Fire up the app in standalone or UnityBox as normal.
- As is the norm these days, some unit and most functional tests are failing. This is present on `master`. These changes add no additional failures.
## Review
- @kurzn
- @higs4281
- either @mistergone or @niqjohnson 
## Example screenshot

screen shot 2016-01-26 at 3 41 45 pm
## Todos

My checklist of current dynamic content changes:
- [x] School name
- [x] Remaining cost
- [x] Total debt
- [x] ~~Private loan fee example~~ (Talked with @kurzn about this, and an example shouldn't be dynamic. No dynamicism for this bullet!)
- [x] Payment plan loan rate
- [ ] Payment plan loan due date
- [ ] Borrowing more than needed (getting to negative remaining cost, as fin aid offers often do)
- [ ] Years attending school (converted into words, for example 'two' instead of '2')
- [ ] Job placement rate
- [ ] Regional or national indication for BLS data
- [ ] National total debt average
- [ ] All the debt burden calculations
  - [ ] "On average, students from your school graduate with $XX of debt"
  - [ ] "compared with a national average of $XX"
  - [ ] projected salary/year
  - [ ] projected salary/month
  - [ ] projected loan payment
  - [ ] debt payment as percentage of income
## Checklist


niqjohnson:
Looking great! I noticed a few things:
- None of the dollar amounts are formatted quite right yet (missing commas, not rounded to the nearest dollar). I assume that all the amounts that get put in by the JS are going to get the right formatting when that work is complete, but what about the amounts Django puts in (`{{school.median_total_debt}}`)?
- Speaking of `{{school.median_total_debt}}`, is that the right number to use there? I think the rest of the numbers in step 2 (i.e. all the bar graphs) are supposed to use program-level data if it's available, then fall back to school-level data if the program data isn't there. So should this be `{{program.median_student_loan_completers}}` or `{{program.mean_student_loan_completers}}` instead?
- Looks like the dollar sign is missing from the big question (or is that going to come back with the JS formatting work, too?).
- Do we need to functional test that the values Django is putting in are the ones we're expecting?
- This one doesn't need to hold up this PR, but we should figure out a general strategy for what we're going to show if any of the numbers isn't available (i.e. should we leave the "[XX]" in the HTML?).

marteki:
> - None of the dollar amounts are formatted quite right yet (missing commas, not rounded to the nearest dollar). I assume that all the amounts that get put in by the JS are going to get the right formatting when that work is complete, but what about the amounts Django puts in (`{{school.median_total_debt}}`)?
> - Speaking of `{{school.median_total_debt}}`, is that the right number to use there? I think the rest of the numbers in step 2 (i.e. all the bar graphs) are supposed to use program-level data if it's available, then fall back to school-level data if the program data isn't there. So should this be `{{program.median_student_loan_completers}}` or `{{program.mean_student_loan_completers}}` instead?

Talked with @higs4281 about this. He would be able to deliver currency-formatted strings for the variables we're not doing any math on. But to your point, due to having to check in the JS as to whether program level data is available, we should just be able to use the JS format-usd conversions to pass the right strings. I then talked with @mistergone, and we'll just probably pass everything we need through the JS formatting.

I'll make changes to this PR to set that up to be easier, and there will be a different PR specifically for the formatting.

> - Looks like the dollar sign is missing from the big question (or is that going to come back with the JS formatting work, too?).

That was me making a mistake. :fearful: Oops.

> - Do we need to functional test that the values Django is putting in are the ones we're expecting?

:stuck_out_tongue: I think we both know the answer to that question.

> - This one doesn't need to hold up this PR, but we should figure out a general strategy for what we're going to show if any of the numbers isn't available (i.e. should we leave the "[XX]" in the HTML?).

Agreed, we need to define this. There should only be a few values where a blank/null value would be sent. I'll work with Bill to come up with that list.

My to-do before asking for another review of this PR:
- [x] Add the dollar sign to the big question
- [x] Tag the Django values with elements to set up JS currency formatting and functional testing
- [x] Add functional tests for the Django values

niqjohnson:
Looks like a good start.

I noticed job placement rate ("Keep in mind your school has a job placement rate of 0% for students who graduate...") and median graduate debt ("On average, students from your school graduate with $0 of debt...") are both coming back as 0. It looks like both are non-zero in `window.schoolData` and `window.programData` for our sample school, so I'm not sure if the zero values are an error or are something that's going to get addressed later.

marteki:
Thanks, Niq. Residuals from when i was trying to take on "not quite all the things, but close".

niqjohnson:
:+1:

Looks good to me now.

mistergone:
Looks like there'll be a possible merge conflict in `dd-functional-settlement-spec.js` but it should be easy to fix. Otherwise, looks good! :+1:

marteki:
Chuck, I'm not seeing any merge conflicts on this. Going to send it through in order to prevent the likelihood that one develops. :smile:

PR #83, opened 2016-01-27
add url to school payload

higs4281:
We have website data for most of our schools, although it's usually just the home page.
## Additions
- adds "url" attribute to the schoolData object when worksheet is rendered
## Review
- @niqjohnson


niqjohnson:
:+1:

URL is getting returned along with the rest of the school data. Thanks!

PR #84, opened 2016-01-27
Add link to feedback form and tests

niqjohnson:
Adds link to the feedback form (and tests)
## Additions
- Link to feedback form from the "Tell us how" button at the bottom of the page
- Functional tests
## Changes
- The "Tell us how" `button` is now an `a`
## Testing
- To test, pull in the `feedback-link` branch and run `gulp`. Fire up the app in standalone or UnityBox as normal. The "Tell us how" button should open the feedback from in a new tab/window.
- I've tested this in IE8/Win7, IE9/Win7, IE10/Win7, IE11/Win7, Firefox/Win7, Chrome/Mac, Safari/Mac, Mobile Safari/iPhone6, and everything seems to be looking good
- Note that this branch is failing some unit and functional tests, just like `master` is right now. The changes here should add no additional failures.
## Review
- @marteki, @caheberer, @kurzn: I followed how the current tool works and am opening the feedback form in a new tab/window. Does that make sense, design-wise? (Actually, the current tool opens the form in a pop-up, which gets blocked by some browsers, so I'm using a plain old new tab/window with `target="_blank"`)
- @marteki, @mistergone: Do the new functional tests look OK? I added them in their own spec to prevent any merge conflicts with the other testing stuff that's being worked on; I can move them elsewhere if it makes more sense. Also, I'm pretty sure I'm testing for the presence of a new tab correctly, but let me know if that seems off to you.
- @higs4281: Is `href="/{{url_root}}/understanding-your-financial-aid-offer/feedback"` the Django-approved way to link to the feedback page?
## Checklist
- [ ] Passes all existing automated tests

### DIFF
--- a/test/functional/settlementAidOfferPage.js
+++ b/test/functional/settlementAidOfferPage.js
@@ -433,8 +433,8 @@ settlementAidOfferPage.prototype = Object.create({}, {
       }
     },
     followFeedbackLink: {
-      value: function( otherexpensesamount ) {
-        return this.feedbackLink.click();
+      value: function() {
+        this.feedbackLink.click();
       }
     }
 
--- a/test/functional/settlementAidOfferPage.js
+++ b/test/functional/settlementAidOfferPage.js
@@ -425,6 +425,17 @@ settlementAidOfferPage.prototype = Object.create({}, {
       get: function() {
         return element( by.css( '.get-options' ) );
       }
+    },
+    //Feedback
+    feedbackLink: {
+      get: function() {
+        return element( by.css( '.feedback .btn' ) );
+      }
+    },
+    followFeedbackLink: {
+      value: function( otherexpensesamount ) {
+        return this.feedbackLink.click();
+      }
     }
 
 } );
--- a/paying_for_college/templates/worksheet.html
+++ b/paying_for_college/templates/worksheet.html
@@ -2547,10 +2547,12 @@
                             </p>
                         </div>
                     </div>
-                    <button class="btn btn__full-small" type="button"
-                    title="Give feedback on this tool">
+                    <a class="btn btn__full-small"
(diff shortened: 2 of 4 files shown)


higs4281:
@niqjohnson not Django-approved, per se, but it allows us to switch from paying-for-college2 to the real thing by changing one view value rather than chasing down all the places it might be, when the time comes.

marteki:
Working in standalone Mac/Chrome and on my phone's Android Chrome Beta browser! I think the new tab interaction is the right choice.

I only saw one thing in the test spec to improve. I'll add the diff comment about it now. Keeping it in its own spec seems wise.

test/functional/settlementAidOfferPage.js
```diff
@@ -425,6 +425,17 @@ settlementAidOfferPage.prototype = Object.create({}, {
...
       }
+    },
+    //Feedback
+    feedbackLink: {
+      get: function() {
+        return element( by.css( '.feedback .btn' ) );
+      }
+    },
+    followFeedbackLink: {
+      value: function( otherexpensesamount ) {
```
marteki: Can the `followFeedbackLink` function be similar to `confirmVerification` (line 23)? There's a unneeded parameter being passed, and it doesn't need to return anything.

niqjohnson:
Updated `followFeedbackLink` to get rid of the extra stuff (I should probably look at what I'm copying an pasting before I copy and paste it...).

Also added a few lines to the tests to close the new tab that was opened in case more tests need to run in the original tab after.

marteki:
Looking and working snazzily, Niqorama! :+1: from me.

higs4281:
:+1: works beautifully on unitybox, and the feedback is saved to the database. I had trouble on the last step on standalone, but that's not surprising and not a concern.

PR #93, opened 2016-02-03
Add link to College Scorecard search

niqjohnson:
Adds a link to a College Scorecard search for schools that offer similar programs as the one selected and are within 50 miles of the selected school's ZIP.
## Additions
- Utility function, `constructScorecardSearch`, to create the Scorecard search query
- Unit tests for `constructScorecardSearch`
- New function in `links-view`, `setScorecardSearch`, to append any query returned by `constructScorecardSearch` to the base Scorecard URL
- Functional tests
- 600 ms delay to functional tests that open and close tabs to account for failures caused by the test running before the tab was ready
## Testing

To test, pull in the `scorecard-link` branch and run `gulp`. Fire up the app in standalone or UnityBox as normal. You can test with our usual test URL:

```
?iped=408039&pid=981&oid=f38283b5b7c939a058889f997949efa566c616c5&tuit=38976&hous=3000&book=650&tran=500&othr=500&pelg=1500&schg=2000&stag=2000&othg=100&ta=3000&mta=3000&gib=3000&wkst=3000&parl=10000&perl=3000&subl=15000&unsl=2000&ppl=1000&gpl=1000&prvl=3000&prvi=4.55&insl=3000&insi=4.55
```

You can also try testing with other IPED IDs to see that (a) the ZIP in the search changes and (b) there's no program in the search (since we don't have program data for those schools). Like Yale (130794), KU (155317), or USC (123961).
### Other testing notes
- The link should open the Scorecard search in a new tab
- I've tested this in IE8/Win7, IE9/Win7, IE10/Win7, IE11/Win7, Firefox/Win7, Chrome/Mac, Safari/Mac, Mobile Safari/iPhone6, and everything seems to be looking good with these caveats: 
  - In IE8, there's an unrelated JS error (present in `master`, too) preventing any of the page's JS from executing
  - In IE9, the link will open correctly, but College Scorecard displays an "update your browser" error and shows no results
  - In IE10, the link will open correctly, but College Scorecard shows no results (it only supports the latest IE, it seems)
  - In IE11 in Sauce Labs, **sometimes** College Scorecard shows an "Error: Forbidden" message; I have no idea why
- This branch is currently failing the same unit and functional tests as `master` is right now. The changes here should add no additional failures.
## Review
- @marteki
- @higs4281: Will CIP codes like "01.0803" always be formatted with the leading zero? The JS is translating CIP code to a general program by taking the first two characters of the CIP code, so if codes like "01.0803" could come through as "1.0803", I can make a change to check for the leading zero.
- @mistergone: Does all the JS structure look OK?
## Screenshots

scorecard-link
## Checklist
- [ ] Passes all existing automated tests

### DIFF
--- a/paying_for_college/templates/worksheet.html
+++ b/paying_for_college/templates/worksheet.html
@@ -2425,9 +2425,14 @@
                                 Review your total cost of attendance and
                                 determine if a less expensive school option
                                 (such as a community college) might also meet
-                                your educational needs. <a href="#">See all
-                                schools</a> that offer the same type of
-                                program in your area.
+                                your educational needs.
+                                <a href="https://collegescorecard.ed.gov/"
+                                class="scorecard-link"
+                                target="_blank">
+                                    See all schools
+                                </a>
+                                that offer the same type of program in your
+                                area.
                             </p>
                         </section>
                         <section class="option option__take-action column-well
--- a/src/disclosures/js/utils/construct-scorecard-search.js
+++ b/src/disclosures/js/utils/construct-scorecard-search.js
@@ -10,7 +10,7 @@
  */
 function constructScorecardSearch( pcip, zip, radius ) {
   var searchParameters = [],
-      searchRadius = radius || '25',
+      searchRadius = radius || '50',
       pcipData = {
         '01': {
           label:  'Agriculture, Agriculture Operations, and Related Sciences',
@@ -175,7 +175,7 @@ function constructScorecardSearch( pcip, zip, radius ) {
     searchParameters.push( 'zip=' + zip );
     searchParameters.push( 'distance=' + searchRadius );
   }
-  return '/search/?' + searchParameters.join( '&amp;' );
+  return 'search/?' + searchParameters.join( '&' );
 }
 
 module.exports = constructScorecardSearch;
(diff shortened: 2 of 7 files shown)


higs4281:
@niqjohnson we store cip and soc codes as strings, so we won't have a truncation issue, as long as the school delivers codes according to our spec, which calls for a string.

higs4281:
:+1:   works like a dream for all but ... Puerto Rico (243197), which has become our PR acid test. 
This is all my fault for telling @marteki about the [Tarzans](https://en.wikipedia.org/wiki/Tarz%C3%A1n_%28mascot%29).

We have the right ZIP, but scorecard apparently doesn't like Puerto Rico double-0 ZIPs. I'm not sure we need to worry about this edge case on scorecard.

niqjohnson:
@higs4281: Great catch. It seems Scorecard will take double-zero ZIPs without complaint, but it fails when it gets a double-zero ZIP **and** a search radius.

I agree with you—this seems like a Scorecard bug that we shouldn't try to work around in our app (I will file an issue in the Scorecard repo, though).

marteki:
Working correctly for the following options: program data available (408039), retiree friendly (407009), alma maters (215062), Oak Park local (149532), and the distant Northlands (102553). :dancer:

Too cool for school. :+1:

PR #98, opened 2016-02-08
Bring the school data (and when available, program data) into the JS model/view area

marteki:
Make it easier for the financial model and financial/metric views to use the school and program data. Yes, this means automatic populating based on school or program data.

This PR just sets up the majority of the model/view stuff up, and only does auto population of the program length selector.
## Additions
- A new model for the school, which holds school and program data
- Populates the program length dropdown if a suggested program length is provided by the school
## Testing
- To test: pull in the `gimme-data` branch and run gulp. Fire up the app in standalone or UnityBox as normal.
- The program length selector should auto-populate for our test ipeds value of 408039, but not for any school without program data. 
- I added one functional test to confirm that the program length selector was being populated when it should be. All tests should pass. :tada: (Edit: except for the still failing Part 2 tests, of course. :blush: )
## Review
- @mistergone @niqjohnson @higs4281 
## Todos
- Populate all the rest of the content that relies on school or program data
- Bar graphs can start to be worked on? :smile: 
## Checklist


marteki:
And it turns out that I _did_ break a former functional test. Fixing it now.

Please hold. Your review is important to us.

niqjohnson:
Niiiiiice! Couple of things that I noticed:
- Setting the program length in the UI with `this.$programLength.val( getSchoolValues.getProgramLength() );` doesn't seem to change the program length used in the calculations ("Loans for [XX] years (program length)" is still 4 times the first-year amount for me when I load the page). Not sure if that's in scope for this PR or not. If not, we should add an issue to get that working to make sure it doesn't fall through the cracks.
- I'm probably the last person you want commenting on code, but `getGradRate` in `get-school-values` jumped out at me since it's returning the whole `schoolValues` object instead of a single value like the rest of the getters there. Feel free to totally ignore this one because, again, last person you want commenting on your code here :smile:

marteki:
> Setting the program length in the UI with this.$programLength.val( getSchoolValues.getProgramLength() ); doesn't seem to change the program length used in the calculations ("Loans for [XX] years (program length)" is still 4 times the first-year amount for me when I load the page). Not sure if that's in scope for this PR or not. If not, we should add an issue to get that working to make sure it doesn't fall through the cracks.

In scope for this PR, as it breaks an old functional test. Both @mistergone and I narrowed down what is causing the breakage (but not how to fix it), so I'm searching the larger world for insight.

> I'm probably the last person you want commenting on code, but getGradRate in get-school-values jumped out at me since it's returning the whole schoolValues object instead of a single value like the rest of the getters there. Feel free to totally ignore this one because, again, last person you want commenting on your code here :smile:

I'm not really the person you want writing your code either! :stuck_out_tongue_closed_eyes: I accept all critique, like a good designer.

higs4281:
:+1: loads program length for me, with 8 test failures, if that's what you're expecting.

marteki:
- Program length setting behaves correctly now.
- Should be down to the six known Part 2 failures.

@mistergone  and @higs4281, thoughts on the `getGradRate` piece that @niqjohnson called out earlier?

niqjohnson:
:+1: 

Program length is working as expected on page load. Nice!

I did notice that `updateView` is now getting called twice in a row in `financialView.init`: once when the `change` is triggered and then again right after that. I can't notice any slowdown when the page loads, though, so I'd vote good to merge and refactor later as needed.

higs4281:
> thoughts on the getGradRate piece

I think I follow how you're passing the values object around to pick up various attributes. No objection here if it's doing the job.

One thought about program length for non-settlement students: the schoolData object has 3 values that tell wether a school offers 2-year, 4-year or graduate programs:

offerAA: "Yes"
offerBA: "Yes"
offerGrad: "Yes"

If we don't want a blank starting point for non-settlement students, we could derive an initial value for program length, say  2 if AA is all that's offered; 4 otherwise? just an option; not a concern for this PR

note: initial comment had an incorrect quesiton about program length.

:+1:

saintsoup52:
In terms of which grad rate we should be showing... when we are at the school level we should be mirroring what college scorecard does and show the graduation level based on the degree primarily offered (in the API)... "primarily associates" vs "primarily bachelors".

As for the data we have "offerAA" and "offerBA" and "offerGrad" I just extracted that from IPEDS data... we should use the same formula for that.

marteki:
> Regarding program length, shouldn't it be reporting 4 years? our test programData has this value:
> 
> level: "Bachelor's degree"

Very odd. `programData` is also giving `programLength: 24`, which I divided by 12 (with rounding up as needed) to get the length in years.

higs4281:
our program data is fake, so inconsistency between programLength and level isn't surprising.  
another reason to get real program data to work with.

marteki:
Now at seven functional test failures. The added one is:

> A dynamic financial aid disclosure that's required by settlement should properly update when a private loan is modified
>   Message:
>     Failed: No element found using locator: By.cssSelector("[data-private-loan] [data-private-loan_key=\"baseAmount\"]")

Which has a fix in PR #99.

I'm going to merge this, with the following to-do notes to myself/the team:
- Double check PR #99 again, then get it into the code base :stuck_out_tongue: 
- Refactor so that `updateView` isn't getting called twice in a row in `financialView.init`
- Look at determining a starter length for non-program data schools based on lowest offered degree at school?

mistergone:
Going to fix a merge conflict, then look into merging #99

PR #100, opened 2016-02-10
Security

higs4281:
Drops use of "shell=True" in subprocess and uses `yaml.safe_load` to address security concerns.
## Review
- @OrlandoSoto @lfatty


lfatty:
Excellent job @higs4281.

PR #101, opened 2016-02-11
Data tweaks

higs4281:
In our attempt to compare apples, this tries to match national earnings and completion rate values with length of program:
- trying first by program level,
- then trying by school predominant degree,
- then trying by school highest degree,
- then falling back to overall national average as last resort
## Additions
- Adds predominantDegree to the schoolData object.
## Fixes
- Our test program object's level value didn't match its program_length, nor our CSV spec. This changes a value in the database, so the fix requires a loaddata command:
  - standalone: `./manage.py loaddata collegedata.json`
  - UnityBox: `./manage.py loaddata collegedata.json --settings=cfpb_django.settings.no_haystack`
## Test
- The schoolObject should now deliver a `predominantDegree` value, which can differ from the `highestDegree` value.
- You should see different national earnings and completion values for 2- and 4-year programs if the app can detect program length:
  - [brown-mackie (408039)](http://localhost:8000/paying-for-college2/understanding-your-financial-aid-offer/offer/?iped=408039&pid=981&oid=f38283b5b7c939a058889f997949efa566c616c5&tuit=38976&hous=3000&book=650&tran=500&othr=500&pelg=1500&schg=2000&stag=2000&othg=100&ta=3000&mta=3000&gib=3000&wkst=3000&parl=10000&perl=3000&subl=15000&unsl=2000&ppl=1000&gpl=1000&prvl=3000&prvi=4.55&insl=3000&insi=4.55) should have a nationalData object with completionRateMedian: 0.3393 and earningsMedian: 31080 -- values for programs of less than 4 years
  - [yale (130794)](http://localhost:8000/paying-for-college2/understanding-your-financial-aid-offer/offer/?iped=130794) should have a nationalData object with completionRateMedian: 0.4778 and earningsMedian: 42051 -- the 4-year program values
## Review
- @marteki @niqjohnson @mistergone

### DIFF
--- a/paying_for_college/disclosures/scripts/nat_stats.py
+++ b/paying_for_college/disclosures/scripts/nat_stats.py
@@ -14,6 +14,8 @@ FIXTURES_DIR = Path(__file__).ancestor(3)
 NAT_DATA_FILE = '{0}/fixtures/national_stats.json'.format(FIXTURES_DIR)
 BACKUP_FILE = '{0}/fixtures/national_stats_backup.json'.format(FIXTURES_DIR)
 BLS_FILE = '{0}/fixtures/bls_data.json'.format(FIXTURES_DIR)
+LENGTH_MAP = {'earnings': {2: 'median_earnings_l4', 4: 'median_earnings_4'},
+              'completion': {2: 'completion_rate_l4', 4: 'completion_rate_4'}}
 
 
 def get_bls_stats():
@@ -63,7 +65,7 @@ def get_national_stats(update=False):
         return json.loads(f.read())
 
 
-def get_prepped_stats():
+def get_prepped_stats(program_length=None):
     """deliver only the national stats we need for worksheets"""
     full_data = get_national_stats()
     try:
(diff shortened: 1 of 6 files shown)


niqjohnson:
:+1: 

All the items in your test section worked for me in both standalone and UnityBox.

marteki:
Working in standalone and Unitybox, for test schools and others.

Are the national data values able to be changed if the program length values change in the view? I'm thinking about how Yale actually has no degrees with program lengths less than four years. If not, would it be better to just deliver all the national values that depend on program length?

higs4281:
@marteki choosing a new national value would require another api call. 
If the value needs to change based on user input, maybe I should just deliver all national data and let JS choose. was trying to take the load off.

but an argument could be made that it's best to compare a school/program to other schools of the same type, regardless of how many years the student intends to be there.

marteki:
Never mind, didn't see @higs4281's correction that colleges are being compared based on predominant degree, not on what program length is passed.

:+1: Yay!

saintsoup52:
Currently... College Scorecard only uses the national average for graduation and earnings... they don't differentiate between 2 and 4 year schools. We should do that accordingly.... ED is considering going to comparing apples to apples, but not for awhile.

Please don't confuse the national average with the graduation rates themselves... they show the 4 year grad rate for predominantly 4 years schools and less than 4 year grad rate for predominately 2 year and certificate granting schools.

PR #102, opened 2016-02-12
fix typo in BLS data; add source in comments

higs4281:
Fix food cost values for Midwest and South
## Review
- @marteki great catches!


marteki:
:+1:

PR #103, opened 2016-02-16
add validation for OID; now allows only hex characters 0-9 a-f

higs4281:
This adds validation to the oid so malicious strings can't be passed to the page.
the oid now can only contain hex characters 0-9 a-f (case insensitive), so angle brackets or any other letters or characters won't be passed.

It also removes code for saving worksheets.
## Deletions
- Removes urls, views, validators and tests for saving worksheets, which is now outside project scope
## Testing
- Sending an oid with anything but hex characters will return an error message:
  "Offer ID has illegal characters; only 0-9 and a-f are allowed."  
## Review
- @OrlandoSoto @lfatty

### DIFF
--- a/paying_for_college/disclosures/urls.py
+++ b/paying_for_college/disclosures/urls.py
@@ -40,11 +40,11 @@ urlpatterns = [
         SchoolRepresentation.as_view(),
         name='school-json'),
 
-    url(r'^api/worksheet/([1-z0-9-]*).json$',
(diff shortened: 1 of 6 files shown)


OrlandoSoto:
:+1:

OrlandoSoto:
thanks @higs4281 , we'll look at the Checkmarx results to confirm

PR #107, opened 2016-02-18
deliver school median salary value consistently

higs4281:
Makes salary value consistent when there is no value; API will return either a number or null, which will avoid NaN.
## Review
- @mistergone @niqjohnson @marteki any or all; a minor change but should cut down oddities.
## Screenshots

null_value

### DIFF
--- a/paying_for_college/models.py
+++ b/paying_for_college/models.py
@@ -200,7 +200,7 @@ class School(models.Model):
             'highestDegree': self.get_highest_degree(),
             'indicatorGroup': jdata['INDICATORGROUP'],
             'KBYOSS': self.KBYOSS,
-            'medianAnnualPay': str(self.median_annual_pay),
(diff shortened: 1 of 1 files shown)


niqjohnson:
:+1:

The school in your screenshot was returning a `medianAnnualPay` of "None" for me previously; it returned `null` with your changes.

@higs4281: Will schools who report a "0" for the median annual pay have a `null` as their `medianAnnualPay` value now, too?

higs4281:
We don't know what schools will report in program data, but for schoolwide
values, it appears scorecard is not delivering zeros anymore.

0 annual pay for graduates seems highly unlikely as a real value. I think
we can safely report it as no data available.

But the code will pass through whatever is reported.

niqjohnson:
Double :+1: !

PR #109, opened 2016-02-19
provide low and hi average ranges for completions and earnings

higs4281:
Adds 4 values to nationalData object, to provide average range.
## Testing

nationalData should now include these values:
- completionRateMedianHigh
- completionRateMedianLow
- earningsMedianHigh
- earningsMedianLow
- loanDefaultRateLow
- loanDefaultRateHigh
## Review
- @niqjohnson


higs4281:
@niqjohnson You'll need to run loaddata to pull in the new placeholder low and high default constants.

NOTE: I standardized on using "High" in the variables instead of sometimes "Hi"

niqjohnson:
A million :+1:! Thank you!

PR #111, opened 2016-02-22
Grad rate

marteki:
Fixes the values pulled into the financial model for graduation rate.

`completionRate` is the term used at the program level, while `gradRate` is used at the school level. They are pointing at the same type of metric.
## Testing

Six functional tests failing, as on `master`.
## Review
- @niqjohnson


niqjohnson:
:+1:

PR #117, opened 2016-02-23
Display graph notifications

niqjohnson:
Hooks up the notification boxes below each graph to the school and program data to show the correct message
## Additions
- All possible notification messages for each metric
- JS to compare school and national values and show the proper notification message for each graph
- Unit and functional tests
## Testing
- To test, pull in the `graph-notifications` branch and run `gulp`. Fire up the app in standalone or UnityBox as normal.
- Try testing with our usual example school, good old `iped=408039&pid=981`. The notifications should look like the screenshot below.
- <strike>Try testing with your favorite outlier schools. For instance, `iped=420866` has no reported graduation rate and a reported average salary that's within the "about the same as average" range. Or `iped=234155` has a default rate that's exactly the same as the national average.</strike> **Graphs for schools without program data aren't working right now; see #120**
- Note that the debt burden notification is not working yet, so you shouldn't see any notification box under the debt burden equation yet.
- Also note that there's a known issue (#114) that's preventing any schools without program data from showing the right graduation rate. That issue is being corrected in an upcoming PR.
- Also also note that there's another known issue preventing JS from executing in IE, so no notifications will appear in IE until that issue is resolved.
## Review
- @marteki 
- @mistergone 
- @higs4281: Throw your favorite outlier schools at this and let me know if any of them break it
## Screenshots

notifications
## Todos
- Debt burden notification (once the debt burden calculation is coded)
- Fix #114 to display graduation rates for all schools
## Checklist
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
- [x] Passes all existing automated tests
- [x] New functions include new tests
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Visually tested in supported browsers and devices 
- [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

PR #118, opened 2016-02-23
Api Refactor

mistergone:
OMG, many change. Such commit. So diff.
## Additions
- AJAX requests for school data and program data
- A method of publishing many financial model changes at once by passing an object to a dispatcher (in `publish-update.js`)
## Removals
- `window` object requests for school and program data 
## Changes
- Changed app to adapt to new source of school and program data
- Changed some tests for correct debt values (`institutionalLoanRate` was not being imported from the URL properly, and I had accidentally typed the same incorrect rate into the spreadsheet)
## Testing
- You shouldn't need to `npm install` or anything, but you might as well!
## Review
- @marteki @niqjohnson @higs4281 
## Todos
- Oops, I left in some commented-out old code. Ought to remove that.
## Checklist
- [ ] Code can be automatically merged (no conflicts)
- [ ] Visually tested in supported browsers and devices 
## Special Thanks
- @niqjohnson, for closing a PR that would've made more conflicts. :smile: 
## Animated GIF

When I refactor code, I'm like...
groucho-dance

### DIFF
--- a/src/disclosures/js/index.js
+++ b/src/disclosures/js/index.js
@@ -3,9 +3,11 @@
 var fetch = require( './dispatchers/get-api-values' );
 var financialModel = require( './models/financial-model' );
 var schoolModel = require( './models/school-model' );
+var getSchoolValues = require( './dispatchers/get-school-values');
 var financialView = require( './views/financial-view' );
 var metricView = require( './views/metric-view' );
 var linksView = require( './views/links-view' );
+
 
 require( './utils/nemo' );
 require( './utils/nemo-shim' );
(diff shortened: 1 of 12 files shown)


src/disclosures/js/index.js
```diff
@@ -3,10 +3,13 @@
 var fetch = require( './dispatchers/get-api-values' );
 var financialModel = require( './models/financial-model' );
 var schoolModel = require( './models/school-model' );
+var getModelValues = require( './dispatchers/get-model-values' );
+var getUrlValues = require( './dispatchers/get-url-values' );
```
marteki, line 7: I think the PR is missing the `get-url-values` file.

mistergone, line 7: Yep, I realized I forgot it in the shower this morning, and I was hoping to sneak it in. Coming soon!

niqjohnson:
I think this might be a known issue, but things broke down for me when testing for a school without program data (i.e. any non-settlement school that either doesn't have a `pid` in the URL or has a `pid` that doesn't return any program data).

Is figuring that out out of scope for this PR? The big effect this issue was having in my testing is I can't verify if graphs are working right for schools without program data.

mistergone:
@niqjohnson - I would call that "out of scope" if only to try to keep this PR from languishing.

Can you maybe file an issue that describes what the expected behavior would be, and I'll work on it?

niqjohnson:
Cool, issue filed: #120 

Everything seems to be working for me for our test school with program data, so :+1: there. Nice!

If this PR gets merged before #117, I'll add a note there that testing the graphs without program data won't work until #120 is closed.

mistergone:
Merging this now...

PR #138, opened 2016-03-01
[HOTFIX] Make adjustments to what content was intending to express about debt repayment

marteki:
Need to make a hotfix adjustment for usability testing, since the values passing through weren't the right ones for the content and user experience.
## Removals
- Got rid of a confusing paragraph in the debt burden area.
## Changes
- The "Do you feel..." question now references the total repayment debt value after interest and fees.
## Testing
- It's on the `fix-debt-value` branch. Give it a whirl. I'm looking to commit today, so it can get onto build for testing by tomorrow.
## Review
- Already reviewed by @kurzn 
- One of @niqjohnson @higs4281 @mistergone 
## Checklist
- [x] CHANGELOG has been updated

PR #140, opened 2016-03-03
Get metric national values from financial model

niqjohnson:
Changes graph and debt burden JavaScript to get all values from the updated financial model instead of the `window` objects
## Removals
- Unneeded test for median school/program debt to make sure all tests pass (that value was removed from the page in #138)
## Changes
- All graph and debt burden values are now pulled from the financial model
- Graphs are updated in the callback that fires when someone verifies their offer information instead of in a separate timeout
## Testing
- To test, pull in the `graph-model` branch and run `gulp`. Fire up the app in standalone or UnityBox as normal.
- Load any test school you like. Everything should look the same as before.
- All unit and functional tests should pass
## Review
- @marteki 
- @mistergone
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,8 @@ All notable changes to this project will be documented in this file.
 We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 
 ## Unreleased
--
--
 - Added animation to button interactions
 - Change graphs to get all values from the financial model
--
--
 
 ## 2.0.2
 - Add active class to selected big question button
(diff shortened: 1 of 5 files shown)


mistergone:
:heart: this.

Just a random note - I think pulling this data from the financial model is fine, although I think it also mostly exists in the school model, too. It's worth noting because the financial model is manipulated and updated and changed a whole lot, whereas the school model is not. I don't think this mostly matters now, and it certainly isn't worth changing anything for this reason (because I don't anticipate the manipulation of the financial model to muck up the school values), but it might become useful in future development.

:+1:

marteki:
Thanks for removing the forgotten functional test, Niq!

Working for me on both standalone and UnityBox. :+1:

niqjohnson:
@mistergone: Good point. Should be pretty painless if we need to start getting national values from the school model later on. Until then, merging!

PR #152, opened 2016-05-03
Changed the standalone base header and footer

higs4281:
This mimics the standalone scheme we used for Retirement. 
It picks up static versions of the on-demand header/footer used on cfgov.
It also converts (again) to Python 2.7, Django 1.8 and Elasticsearch
## Additions
- static versions of the on-demand header/footer templates
## Testing
- Fire up with runserver 
## Review

any of @mistergone  @niqjohnson @marteki 
## Todos
- footer icons don't display on some pages; may be a local issue
## Checklist

### DIFF
--- a/paying_for_college/templates/standalone/base_update.html
+++ b/paying_for_college/templates/standalone/base_update.html
@@ -35,28 +35,22 @@
     <meta name="viewport" content="width=1016">
     {% endblock %}
 
-    <!-- Favicons -->
-    <link rel="apple-touch-icon" href="http://www.consumerfinance.gov/wp-content/themes/cfpb_nemo/_/img/apple-touch-icon-precomposed.png">
-    <link rel="apple-touch-icon" size="72x72" href="http://www.consumerfinance.gov/wp-content/themes/cfpb_nemo/_/img/apple-touch-icon-72x72-precomposed.png">
-    <link rel="apple-touch-icon" size="114x114" href="http://www.consumerfinance.gov/wp-content/themes/cfpb_nemo/_/img/apple-touch-icon-114x114-precomposed.png">
-    <link rel="apple-touch-icon" size="144x144" href="http://www.consumerfinance.gov/wp-content/themes/cfpb_nemo/_/img/apple-touch-icon-144x144-precomposed.png">
+    <link rel="shortcut icon" type="image/x-icon" href="http://www.consumerfinance.gov/favicon.ico">
+    <script src="//ajax.googleapis.com/ajax/libs/jquery/1.11.3/jquery.min.js"></script>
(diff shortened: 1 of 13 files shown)


marteki:
Note: I had to run `./local_setup.sh` and revert my standalone to use Django 1.6 before I could get this running.

Look at those beautiful headers! 😍  I think the broken footer icons are due to icon fonts not loading; they're also not loading on the repay student debt tool.

screen shot 2016-05-03 at 12 20 05 pm

But... I noticed some oddness in the guide pages that doesn't seem related. 😞  Check out these page title sizes.

screen shot 2016-05-03 at 12 19 19 pm

screen shot 2016-05-03 at 12 19 31 pm

Seems like they're being reset by `footer.css`, and `guides.css` isn't being triggered last as it should. Take another look at the order of operations for the CSS in the template?

amymok:
@higs4281 The only problem I have is I get an error when running standalone.py (part of the `gulp build`, but probably not related to this PR):

```
gulp-notify: [Compile Error] Error in plugin 'gulp-less'
Message:
    Invalid mapping: {"generated":{"line":1,"column":4613},"source":"/Users/moka/workspace/college-costs/src/vendor/cf-core/src/cf-media-queries.less","original":{"line":1,"column":null},"name":null} in file undefined line no. undefined
Details:
    lineNumber: undefined
    fileName: undefined
```

@niqjohnson @anselmbradford mentioned that we can comment out two lines in gulp/tasks/styles.js:
- 13: `.pipe( $.sourcemaps.init() )`
- 23: `.pipe( $.sourcemaps.write( '.' ) )`

But something we need to figure out to fix them instead of commenting out.

Also I don't have right to merge this PR.  Please sync up with the latest code as well.

paying_for_college/templates/standalone/base_update.html
```diff
@@ -201,187 +130,36 @@
...
 
+
 {% block nemo_js %}
-<script src="http://www.consumerfinance.gov/wp-content/themes/cfpb_nemo/_/js/functions.js"></script>
-<script type="text/javascript" src="http://www.consumerfinance.gov/static/agreements/js/chosen.jquery.js"></script>
+<script src="http://www.consumerfinance.gov/wp-content/themes/cfpb_nemo/_/js/functions.js?ver=all"></script>
+<script type="text/javascript" src="http://www.consumerfinance.gov/static/agreements/js/chosen.jquery.js?ver=all"></script>
 {% endblock %}
-
+<script src="http://www.consumerfinance.gov/static/nemo/_/js/functions.js?ver="></script>
```
anselmbradford: Does this need a value set for `?ver=`?

higs4281: the empty query appears to be a no-op suffix. we could lose it.

PR #160, opened 2016-05-20
fix standalone setup script

higs4281:
migrate should not be specifying the app for a new install


higs4281:
tested with fresh install. merging as hotfix for any newcomers

PR #165, opened 2016-05-23
fix file path for BLS json

higs4281:
When the app runs inside V1 it will need a full file path for data files. This adds the BASE_DIR value to the path.
## Changes
- path for bls_data.json
## Testing
- Should not change anything in standalone (/api/expenses/ should still work)
- in V1 standalone, this enables the endpoint
## Review
- @amymok 
- @marteki this was part of the closed [PR 161](https://github.com/cfpb/college-costs/pull/161/files) that I forgot about


amymok:
Looks good to me, we can merge after @marteki take a look

PR #169, opened 2016-05-25
Fix sync of our migrations with server databases

higs4281:
The deployment issue was that the `offers_perkins` column on the School
model was in our Django 1.8 initial migration, so there is no new column
detected when run on the build server, even with --fake-initial. The new
`settlement_school` column gets detected and created in migration 0002.
The solution is to delete `offers_perkins` from the initial migration
and then run `makemigrations`, which creates a new 0003 migration that
will get the table created in our deployment. it is fervently hoped.
## Additions
- migration 0003 
## Removals
- `offers_perkins` line from initial migration
## Review
- @amymok 
## Checklist


amymok:
looks great, let's try it out

PR #170, opened 2016-05-26
Adjust setup.py to include fixtures

higs4281:
We need to have BLS data and our collegedata fixture available to the
server to support our API and data migration.
## Changes
- Adjusted setup.py to include fixtures 
## Review
- @amymok 
## Checklist


amymok:
Looks good to me

PR #171, opened 2016-05-26
Adjust setup.py to pick up template files

higs4281:
We didn't realize our non-python template files would not be picked up
in the deployment wheel, so we need tpo included them in the `package_data`
list.
## Changes
- Adjust setup.py to catch our templates
## Review
- @amymok 
## Checklist


amymok:
Looks great!

PR #176, opened 2016-06-06
Api command

higs4281:
Add a manage.py command for updating college data.
## Additions
- `update_via_api` command and related test
## Testing
- `./manage.py update_via_api` will launch an update job that will attempt to update data for every school in our database. We would normally run this command only once a year, or if we need to add fields to our data.
- the command takes one optional parameter, `school_id`, which will accept a school ID as an argument value and limit the update to just that school.
- You can test this locally without fear, because it will only touch your local database, and probably won't change anything anyway because this update has already been run for our data. If you launch the script and it starts getting results, you can cancel it with `ctrl-c`.
## Review
- @amymok and anyone else who wants to give it a spin.
## Checklist


amymok:
looks good!

PR #179, opened 2016-06-09
10/25 Year Loan toggle added

mistergone:
This PR adds the 10/25 year loan term toggle to the appropriate metrics sections. I also went down the rabbit hole a bit and refactored `metrics-view`. **Note**: The styles are _mostly_ correct, but there are slight differences between the final design and the current look of the app.
## Additions
- 'How much will I pay per month for all of my student loans?' section
- Radio buttons in two sidebars of Step 2
- Handlers which update various values based on a 10 or 25-year loan term
## Changes
- `metric-view` refactored
  - Added an object property, `metrics`, which stores important numbers for later use
  - Refactored methods to use fewer parameters
  - Removed and/or changed the way several methods work
## Testing
- The app passes both existing and new functional and unit tests. (See the README for how to test.)
## Review
- @marteki and @higs4281 
- @niqjohnson - If you have some time, I'd love for you to check this out. I rewrote a lot of `metrics-view`, and if you can check whether I missed something, that'd be great!
## Screenshots

screen shot 2016-06-09 at 11 22 11 am
## Checklist
## Animated gif

**"Really, Chuck. You think now is a good time to refactor?"**

spock-eye-raise

### DIFF
--- a/src/disclosures/js/models/school-model.js
+++ b/src/disclosures/js/models/school-model.js
@@ -18,7 +18,8 @@ var schoolModel = {
   processAPIData: function( values ) {
     values.jobRate = values.jobRate || '';
     values.programLength /= 12;
-    values.medianSalary = values.salary || values.medianAnnualPay;
+    values.medianSalary = values.salary || values.medianAnnualPay
+      || values.earningsMedian;
     values.monthlySalary = Math.round( Number( values.medianSalary ) / 12 );
     values.medianSchoolDebt = values.medianStudentLoanCompleters ||
       values.medianTotalDebt;
--- a/src/disclosures/js/models/school-model.js
+++ b/src/disclosures/js/models/school-model.js
@@ -18,8 +18,8 @@ var schoolModel = {
   processAPIData: function( values ) {
     values.jobRate = values.jobRate || '';
     values.programLength /= 12;
-    values.medianSalary = values.salary || values.medianAnnualPay
-      || values.earningsMedian;
+    values.medianSalary = values.salary || values.medianAnnualPay ||
+      values.earningsMedian;
     values.monthlySalary = Math.round( Number( values.medianSalary ) / 12 );
     values.medianSchoolDebt = values.medianStudentLoanCompleters ||
       values.medianTotalDebt;
--- a/src/disclosures/js/views/metric-view.js
+++ b/src/disclosures/js/views/metric-view.js
@@ -161,8 +161,6 @@ var metricView = {
   /**
    * Sets text of each point on a bar graph (or a class if a point is missing)
    * @param {object} $graph jQuery object of the graph containing the points
-   * @param {string} schoolText Text of the graph's school point
-   * @param {string} nationalText Text of the graph's school point
    */
   setGraphValues: function( $graph ) {
     var $school = $graph.find( '[data-bar-graph_number="you"]' ),
@@ -184,10 +182,6 @@ var metricView = {
   /**
    * Sets the position of each point on a bar graph
    * @param {object} $graph jQuery object of the graph containing the points
-   * @param {number|NaN} schoolValue Value reported by the school
-   * @param {number|NaN} nationalValue Average national value
-   * @param {object} $school jQuery object of the graph's school point
-   * @param {object} $national jQuery object of the graph's national point
    */
   setGraphPositions: function( $graph ) {
     // schoolValue, nationalValue, $school, $national
--- a/src/disclosures/js/views/metric-view.js
+++ b/src/disclosures/js/views/metric-view.js
@@ -14,6 +14,8 @@ var metricView = {
     var settlementStatus =
       Boolean( getSchool.values().settlementSchool ) || false;
     this.updateGraphs( values, settlementStatus );
+    this.toggleListener();
+    this.updateMonthlyPayment();
     // updateDebtBurdenDisplay is called in financialView.updateView, not here,
     // since the debt burden needs to refresh when loan amounts are modified
   },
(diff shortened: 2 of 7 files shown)


src/disclosures/js/models/school-model.js
```diff
@@ -18,7 +18,8 @@ var schoolModel = {
   processAPIData: function( values ) {
     values.jobRate = values.jobRate || '';
     values.programLength /= 12;
-    values.medianSalary = values.salary || values.medianAnnualPay;
+    values.medianSalary = values.salary || values.medianAnnualPay ||
```
marteki, line 21: All the other lines in this sections seem to be first looking for program level data, then school level data. This one is comparing three values.

Can you talk a little more about what's now going on for the medianSalary value, or point me at the documentation about it? The variable names are so similar, and it's hard to know where each datapoint is coming from or represents. 😕

src/disclosures/js/views/metric-view.js
```diff
@@ -5,94 +5,113 @@ var getSchool = require( '../dispatchers/get-school-values' );
 var formatUSD = require( 'format-usd' );
 
 var metricView = {
+  metrics: {
+    debtBurden: {
+      school: NaN,
```
marteki, line 10: I have a few guesses, but rather than assume, 😄  I'll ask: why are we explicitly initializing these as NaNs?

test/js-unit/metric-view-spec.js
```diff
@@ -4,151 +4,90 @@ var metricView = require( '../../src/disclosures/js/views/metric-view' );
...
+    var notificationClasses = metricView.getNotifications( 'test' );
+    expect( notificationClasses ).to.equal(
+      'cf-notification metric_notification__worse cf-notification__error'
+    );
+  });
+
+  it( 'gives the proper notification classes when min and max ' +
+       'values are not numbers', function() {
+    metricView.metrics.test.national = 0.5;
+    metricView.metrics.test.school = 0.5;
```
marteki: Why did the school value increase here? Can we change this back to 0.3, to match the previous version of the test (refactoring it rather than testing something new)?

src/disclosures/js/views/metric-view.js
```diff
@@ -107,86 +126,94 @@ var metricView = {
...
-   * national average (from College Scorecard)
-   * @param {string} betterDirection 'higher' or 'lower' depending on whether a
-   * school value higher or lower than the national average is more desirable
+   * @param {number|NaN} metricKey - metric to be checked
    * @returns {string} Classes to add to the notification box
    */
-  getNotificationClasses: function( schoolValue, nationalValue, sameMin,
-    sameMax, betterDirection ) {
-    var notificationClasses = '';
+  getNotifications: function( metricKey ) {
```
marteki, line 219: > gulp lint
>   190:21  warning  Function 'getNotifications' has a complexity of 11

😲 

The code refactor is _increasing_ the cyclomatic complexity of this section, and into the double digits. As much as I enjoy Spinal Tap, I'm pretty uncomfortable with the idea of going to 11. Could we use a different approach in the code to keep from doing that?

src/disclosures/js/views/metric-view.js
mistergone, line 10: I think it's best to assume there's no data available, which means no number (rather than showing an inaccurate default number).

test/js-unit/metric-view-spec.js
mistergone: I don't understand your concern. We should talk about it. :)

src/disclosures/js/views/metric-view.js
mistergone, line 219: The complexity number is basically the number of `if` statements in the function. This function isn't a bunch of scary nested `if` statements, but rather a series of `if - else` statements checking the state of the school and national numbers. We could certainly refactor the code, but I'm not convinced it would make it easier to read or more efficient. I'm open to ideas, though!

higs4281:
Code works for me, but it leaves me with a question:  
When a student chooses the 25-year option at "Calculate your monthly payment," shouldn't that choice propagate to the "Your estimated debt burden" section, and vice versa? Otherwise, you have a possibly inadvertent mismatch.

Other than that, 👍

mistergone:
@higs4281 - The original specification was that each section was independent of the other. If you think it's an issue, we can see how everyone (@kurzn @marteki and perhaps @niqjohnson and @caheberer) feels about the final interactions and revise them if necessary.

higs4281:
👍  agreed -- question is outside scope of this PR.

mistergone:
@marteki - I addressed the complexity issue with some refactors which may not make the whole thing easier to read (it's debatable!) but definitely makes it more atomic, with less individual function complexity. Here's the commit message for details:

> To remove the complexity warning for getNotifications(), I refactored this file a bit. Instead of getNotifications() doing the entire lift for determining the 'standing' (higher, lower, about the same) of a school per metric, I added a function, checkMetrics(), which does a simple set of checks of the school values against the high and low to determine the 'standing' property for that metric. These standings are then used by getNotifications() to assign classes. getNotifications() also checks if any data is missing, and returns the proper classes based on all these checks.
> 
> With these changes, I was able to completely eliminate the complexity warning we were getting for getNotifications().

This commit still passes all existing tests. I added five new tests for the new function, `checkMetrics()`, in `/tests/js-unit/metric-view-spec.js`

marteki:
94dnw

Thank you for looking into the complexity!

:+1: from me for merge.

caheberer:
I'm not sure if now is the time to say that the radio button selection state doesn't match our standards...but it should look like this:

screen shot 2016-06-13 at 4 04 03 pm

But in regards to why I was @'ed. From what I recall, we agreed the toggle wouldn't change numbers in other sections at this time.

mistergone:
@marteki  - I see your Pinkie Pie and raise you Cheese Sandwich

niqjohnson:
Looks like a solid refactor to me. Nice!

Only thing I noticed is that the JSdoc comments for some of the refactored methods (`setGraphValues`, `setGraphPositions`) still have some old `@param` notes in them.

mistergone:
@caheberer - I noted it in the Pull Request description, but you probably missed it. This PR is mostly for the front-end functional changes, so the styles are not 100% to spec yet.

caheberer:
@mistergone 
giphy

mistergone:
@niqjohnson - Good catch! In the refactor, I tried to update JSDocs as I went, but I obviously missed some!

PR #181, opened 2016-06-13
Script to load programs

amymok:
A script to load programs with the agreed-upon data fields.
## Additions
- Added script to load programs
- Added tests for script
- Added management command to call the script
- Added test for command
## Review
- @higs4281


paying_for_college/disclosures/scripts/load_programs.py
```diff
@@ -0,0 +1,156 @@
+from __future__ import print_function
+import os
+
+from csvkit import CSVKitDictReader as cdr
```
higs4281: this one's my fault. as much as I **love** csvkit, I don't think we can justify loading it on the servers. I has some heavy dependencies, including SQLAlchemy.

For our use, importing the standard csv library should be fine, although it won't handle encoding oddities as well:

```
from csv import DictReader as cdr
```

amymok:
@higs4281 This is ready for another review.

higs4281:
👍 thorough!
the requirements are just for standalone, but we might save testers time by removing csvkit from requirements/base.txt

PR #182, opened 2016-06-14
Percentage, private loan fee fixes

mistergone:
This fixes a bug in percent fields and brings in private loan fees from the URL (or Direct Loan rate).
## Additions
- Added `prvf` to query handler
- Added a fix for floating point imprecision showing in the percent input fields
  - Added `round()` function to fix this imprecision
## Testing
- Passes existing functional and unit tests
## Review
- @marteki and @higs4281 and @amymok and anyone who loves floating point errors
- Pull it down, `gulp` it, and check out the `prvf` URL parameter or try to type numbers like `3.666` in percent fields and watch the lack of floating point errors!
## Checklist
## Animated GIF

It seems like a **MEGASHARK** kind of day.

megashark-bites-bridge

### DIFF
--- a/src/disclosures/js/views/financial-view.js
+++ b/src/disclosures/js/views/financial-view.js
@@ -43,6 +43,16 @@ var financialView = {
   },
 
   /**
+   * A better rounding function
+   * @param {number} n - Number to be rounded
+   * @param {number} decimals - Number of decimal places
+   */
+  round: function( n, decimals ) {
+    var number = n + 'e' + decimals;
+    return Number( Math.round( number )+ 'e-' + decimals );
+  },
+
+  /**
    * Helper function that updates the value or text of an element
    * @param {object} $ele - jQuery object of the element to update
    * @param {number|string} value - The new value
@@ -71,7 +81,7 @@ var financialView = {
     $percents.not( '#' + financialView.currentInput ).each( function() {
       var $ele = $( this ),
           name = $ele.attr( 'data-financial' ),
-          value = values[name] * 100;
+          value = financialView.round( values[name] * 100, 2 );
       financialView.updateElement( $ele, value, false );
     } );
   },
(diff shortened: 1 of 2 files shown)


src/disclosures/js/views/financial-view.js
```diff
@@ -43,6 +43,16 @@ var financialView = {
   },
 
   /**
+   * A better rounding function
+   * @param {number} n - Number to be rounded
+   * @param {number} decimals - Number of decimal places
```
anselmbradford, line 48: Include something like `@returns {number} The rounded value.`

marteki:
With `prvf` in URL of 2.1:

screen shot 2016-06-14 at 5 12 57 pm

With `prvf` in URL of 5:

screen shot 2016-06-14 at 5 13 17 pm

With no `prvf` passed in the URL:

screen shot 2016-06-14 at 5 13 31 pm

Hmm. What's the Direct Loan origination fee percentage, again? Let me check our constants API:

> "DLOriginationFee": "1.01073"

**I assume that only the amount after the decimal is the percent to be used.** If not, this isn't quite right. If so, then huzzah!

And decimals seem to be working much better in the rate boxes!

:+1:

mistergone:
@ans - Good catch, thanks!

@marteki - You have discovered a secret inconsistency that occasionally annoys me - almost all our other fees are written as decimals, but we take the Direct loan fee as a coefficient, if you will (basically, a 1% fee is turned into a 1.01 coefficient). I don't like that inconsistency, but it would be pretty annoying to extricate at this point. Anyway, the code subtracts 1 from the fee coefficient, which makes it the decimal equivalent of the percent. Makes sense, right? :wink:

marteki:
Makes sense; it's what I figured was going on. And now that you pointed it out, I totally see the subtraction of the 1 from the passed fee value.

higs4281:
👍 welcome to our party, prvf.

mistergone:
JSDoc omission added, merging this!

merge4

saintsoup52:
The default private student loan fee (when not populated) should be 0% based on feedback from RMR

kurzn:
@saintsoup52  The direction we had always gotten from you (via the Word doc commentary) has been to default to the Grad Plus loan rate. 

image

Is this a new requirement? If so, it wasn't ever communicated to the team.

saintsoup52:
This is my bad... RMR told me they couldn't find but one example of a lender charging a fee for a private loan.

the good news is this should be a constant that is pulled in like all the other loan constants (e.g., PLUS loan rates). Is that right? @higs4281

mistergone:
@saintsoup52 - Any private loan fees can be passed through the URL. We'll remove the fallback to Direct loan fees and fallback to 0 when that value is absent.

higs4281:
We seem to be conflating fees and rates. 

Our constants deliver a fallback value for private loan rates, but not a fallback fee value. From Patrick's comment, the fallback fee value should be 0 anyway, since RMR says only one lender charges a fee.

My understanding is this -- please correct me if I'm off base: 
- fee is an up-front charge, expressed as a percentage point, often called "points" on a loan
- rate is the ongoing interest rate, expressed as a percentage rate, charged until the loan is paid off.

saintsoup52:
Sorry to sow confusion here... My question was a tactical question...

Currently the default private loan interest rate should be 7.9% and the loan fee should be 0%, however in the future those numbers might change (e.g., if we do a new private student loan study).

Would it be easier to update these values in the code? Or should we just add these default values to the list of constants we regularly update (e.g., the subsidized federal student loan interest rate).

higs4281:
We currently deliver the 7.9% private loan rate in our constants:
http://build.consumerfinance.gov/paying-for-college2/understanding-your-financial-aid-offer/api/constants/
We don't deliver the 0 default for private loan fee, but it would be easy to add if that would future-proof us

saintsoup52:
I like the idea of "future-proofing" this way.

PR #190, opened 2016-06-23
Remove 6 fields from school json payload

higs4281:
The deleted fields are not being used or can be derived from current data from
the Ed API.
## Removals

These fields are dropped from the school payload:
- KBYOSS
- avg_net_price
- indicatorGroup
- offerAA
- offerBA
- offerGrad
## Testing
- hit a school endpoint, and the values should be gone.

```
http://localhost:8000/paying-for-college2/understanding-your-financial-aid-offer/api/school/130794/
```

yale_json
## Review
- @amymok @marteki @mistergone 
## Checklist
- [x] Data dictionary and CHANGELOG updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ All notable changes to this project will be documented in this file.
 We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 
 ## Unreleased
+- removed 6 unneeded fields from school json payload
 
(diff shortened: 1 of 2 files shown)


marteki:
The six fields aren't in the endpoint any more.

The tool still loads and works as it does on `master`.

:+1:

amymok:
Looks good to me as well

PR #193, opened 2016-06-24
Change require fields for program data

amymok:
Changed the requirement of which program data fields are mandatory
## Changes
- Made program code required
- Made program name not required
## Review
- @higs4281 
## Todos
- Need to think about if we should log information if there were too many empty fields.  If so, how as well.


higs4281:
👍 we can try this out with new data

PR #194, opened 2016-06-27
Add logging and messaging to Notification model

higs4281:
We need to track and retry notification failures, so we're adding a log
entry to each Notification record to track its own transmission attempts.
We're also adding a "sent" boolean that is True
if a notification was successful. In this way we can regularly poll the
database to see if there are send failures and retry them or, if they're
getting old, collect them for troubleshooting and for notifying schools.
## Additions
- notifications.py script to resend and gather stale notifications.
- tests to match
## Testing

This is tough to test locally, because you need some notification records locally, resending notifications will fail (EDMC will reject any postings from outside EXT) and sending email for stale notifications requires an email host to be configured. These will be more easily tested in Jenkins once we push to servers.
## Review
- @amymok 
- anyone else who'd like to look it over.
## Checklist
## Note

Since this PR makes a model change, `./manage.py migrate paying_for_college` is needed to keep your local db in sync.
## To do
- create a mangement command to run notifications

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 - Error checking added to program view
 - Error checking added to verification view
 - removed 6 unneeded fields from school json payload
+- Added notifications script and 2 fields to Notification model for errors
 
 ## 2.1.2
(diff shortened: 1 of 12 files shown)


amymok:
looks great

PR #197, opened 2016-06-28
Configure JMeter Properties

amymok:
Currently jmeter creates a jtl file which is in csv format, which causes a number format java exception when running the load test in jenkins.  To prevent this exception to occur, the jtl file needs to be a xml file, and this properties file will help configure that if using performance report plugin.
## Additions
- Added JMeter properties to emit xml in jtl file
## Review
- @higs4281 @OrlandoSoto


higs4281:
👍 looks good

OrlandoSoto:
thanks @amymok !

PR #206, opened 2016-07-05
Include Clothing/Taxes as separate BLS json data

amymok:
Clothing and Taxes are currently not pulled out in our BLS json, now pulling them out.
## Additions
- Added in taxes in BLS json
## Changes
- Changed Clothings from Others category to its own
## Review
- @higs4281


higs4281:
👍 just right.

PR #213, opened 2016-07-13
Refactor script tests to use the mock_open helper

higs4281:
Previous attempts to mock the built-in 'open'function ended in tears.
## Changes
- update_ipeds unit tests refactored
- @amymok (thanks for the tip!)


amymok:
beautiful!

PR #219, opened 2016-07-15
We need to catch rates that are in the wrong form

higs4281:
Our CSV spec failed to explicitly say that rates should be in the form "0.55" instead of "55" 
to represent 55%. This PR catches any rates that are expressed as >1 numbers and converts them.
It also catches an encoding issue that cropped up in the first batch of EDMC programs.
## Additions
- Two helper functions to load_programs
- tests to match
## Review
- @amymok 
## Checklist

### DIFF
--- a/paying_for_college/disclosures/scripts/load_programs.py
+++ b/paying_for_college/disclosures/scripts/load_programs.py
@@ -73,7 +73,7 @@ def read_in_data(filename):
             reader = cdr(f)
             data = [row for row in reader]
     except:
-        return {}
+        return [{}]
     else:
         return data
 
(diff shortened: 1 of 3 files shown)


paying_for_college/disclosures/scripts/load_programs.py
```diff
@@ -89,18 +89,33 @@ def clean_string_as_string(string):
     return clean_str if clean_str.lower() not in NO_DATA_ENTRIES_LOWER else ''
 
 
+def standardize_rate(rate):
+    if rate and float(rate) > 1:
+        return unicode(float(rate)/100)
```
amymok, line 94: Should we use `Decimal` instead of `unicode` here?

paying_for_college/disclosures/scripts/load_programs.py
```diff
@@ -89,18 +89,33 @@ def clean_string_as_string(string):
...
+    else:
+        return rate
```
amymok, line 96: Same question, should we wrap this around like this: `Decimal(rate)`

higs4281, line 96: No need in either case. all the 'clean' commands just return strings, and string is later converted to Decimal on insertion. I just ran upload_programs to check, and all the rates load to Decimals without issue.

amymok, line 96: Sounds good.  All good after you resolve conflicts!

PR #220, opened 2016-07-18
Allow purging of notifications and programs

higs4281:
We need to be able to purge notification and program objects via Jenkins
in order to prepare for live testing by a school. This adds manage.py
commands for doing that, plus a model function for ensuring that
program_length values that are delivered as json are always even.
## Additions
- purge command and purge_objects script to support it
- related tests
- model function to deliver even values for program_lengths
## Testing
- Try wiping programs and notifications from your local db.

```
./manage.py purge notifications
./manage.py purge programs
```
## Review
- @amymok and any purge-happy folk
## Checklist

### DIFF
--- a/paying_for_college/disclosures/scripts/purge_objects.py
+++ b/paying_for_college/disclosures/scripts/purge_objects.py
@@ -1,0 +1,21 @@
+from paying_for_college.models import Notification, Program
+
+error_msg = "Only Notification and Program objects may be purged"
+no_args_msg = ("You must supply an object type to purge, "
+               "either notifications or programs")
+object_map = {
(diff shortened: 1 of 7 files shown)


paying_for_college/disclosures/scripts/purge_objects.py
```diff
@@ -0,0 +1,21 @@
+from paying_for_college.models import Notification, Program
+
+error_msg = "Only Notification and Program objects may be purged"
```
amymok: It would be helpful if we mention using `notifications` and `programs` only as argument.

amymok:
Other than the comments above, everything looks good~

PR #225, opened 2016-07-25
Ipeds fix

higs4281:
This makes our IPEDS downloader check for an existing CSV before downloading and parsing a new one. Since we only need to run once a year, if it runs again there's no need to download raw files again.
## Additions
- Adds a file-exists check to the `read_csv` function
## Changes
- Adjusts tests to match, plus pep8 tweaks
## Review
- @amymok 
## Checklist


amymok:
@higs4281 looks good as well.  but need to resolve conflicts before merging.

PR #232, opened 2016-07-27
Cleans up header - adds quotes to attributes

anselmbradford:
## Changes
- Adds missing quotes to attributes in header HTML.
- Performs general cleanup (removing extraneous whitespace, indenting issues).
## Review
- @marteki 
## Notes
- Changes are too big to view it looks like. You'll have to take my word for it ;) You can view https://github.com/cfpb/college-costs/blob/2653e48620345958771ea727b30bf09c86d5cd09/paying_for_college/templates/standalone/on-demand/header.html and search for `href` and see that all values are quoted now.

### DIFF
--- a/paying_for_college/templates/base.html
+++ b/paying_for_college/templates/base.html
@@ -2,6 +2,6 @@
 <html>
 <head></head>
 <body>
-{% block content %}{% endblock %}    
+{% block content %}{% endblock %}
 </body>
 </html>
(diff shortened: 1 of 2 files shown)


marteki:
Note to @higs4281 and other Python developers, and any coders like me who might have missed it: in HTML, we should quote all attributes.

:+1:

PR #234, opened 2016-07-27
Add a private loan fee value to our constants

higs4281:
Following up on a [PR comment](https://github.com/cfpb/college-costs/pull/182#issuecomment-226844101) by Patrick about future-proofing, this gives us a future home for private loan fees, if they ever become a thing. 
The current value is 0, because we're not aware of any private loans that are charging fee points.
## Additions
- New constant, `privateLoanFee`, with an initial value of 0.
- This affects only our database content -- no code.
## Testing
- After pulling and running `./manage.py loaddata collegedata`, `privateLoanFee` should show up in [constants](http://localhost:8000/paying-for-college2/understanding-your-financial-aid-offer/api/constants/)
## Review
- @saintsoup52 @amymok 
## Screenshots

private_loan_fee
## Checklist
- [x] CHANGELOG updated


amymok:
Tried it, looks good to me as well!

saintsoup52:
I know I am late to the party on this... but thank you for future proofing us!

PR #235, opened 2016-07-28
Add unicode-aware csv handling

higs4281:
The csv reader and writer classes from the csvkit python package  provide 
more reliable handling of encoding issues by reading in unicode by default, 
including encoding options for reading files, and writing out utf-8 by default.
csvkit and its successor, agate, carry some heavy dependencies and 
would be overkill for this app's needs.
## Additions
- csvkit reader and writer classes
## Removals
- dependencies on full csvkit package
## Review
- @amymok 
## Checklist
- [x] CHANGELOG updated

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -8,6 +8,7 @@ omit =
     */python2.7/*
     *setup.py
     */data_sources/*
+    */csvkit/*
 
 [report]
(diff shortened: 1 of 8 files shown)


amymok:
Looks good to me!

PR #237, opened 2016-07-28
Return 200 with messages instead of 400 errors

higs4281:
We need to trap school and program ID errors on the page instead of
in the back end, so we'll now return 200 and render the offer page
with a warning message.

These messages are available at `$('div.verify_content').attr('data-warning')`

| Problem | Message |
| :-- | :-- |
| Missing iped | Warning: URL doesn't contain a school ID |
| Bad iped `xxxxxx` | Warning: No active school found for ID xxxxxx |
| Missing pid | Warning: URL doesn't contain a program code |
| Bad pid `xxx` | Warning: No program could be found for program_code 'xxx' |
| Illegal pid `<script>` | Error: Illegal program code |
| Missing oid | Warning: URL doesn't contain an offer ID |
| Illegal oid `<script>` | Error: Illegal offer ID |
## Changes
- offer view refactored to return 200 and warning messages instead of 400
## Testing
- Try pages with bad or missing school ID or program code or offer ID
- Try pages with illegal characters (such as '<') in the program code or offer ID
## Review
- @amymok @mistergone @marteki @anselmbradford 
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 - Update debt summary text for program durations less than one year.
 - Added privateLoanFee constant to the database
 - Added unicode-aware csv module
+- Refactored the offer view to allow bad school, program and offer IDs
 
 ## 2.1.4
 - Added string checking for program codes
--- a/paying_for_college/disclosures/scripts/update_colleges.py
+++ b/paying_for_college/disclosures/scripts/update_colleges.py
@@ -95,7 +95,6 @@ def update(exclude_ids=[], single_school=None):
                         CLOSED += 1
                     if UPDATED is True:
                         UPDATE_COUNT += 1
-                        # school.data_json = json.dumps(data_dict)
                         school.zip5 = fix_zip5(str(school.zip5))
                         school.save()
                 else:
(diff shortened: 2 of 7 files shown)


amymok:
Looks good!

saintsoup52:
I am glad we are focusing on this... though I must say i don't really know what the difference between a 200 and 400 code, I am assuming it just warns the user if something is wrong.

My only concern is the tool should still work if there is no PID or OID in the URL. The includes two different situations:
1. "pid=" -  Where the PID / OID does not appear in the URL at all. 
2. "pid=&" - Where the PID / OID is NULL or absent

We will need to demo this tool to other settlement schools and without their program data we would never be able to show them how their schools looks if have to use a working PID. As for the OID, it would also just be easier to demo the tool if we did not have to include one.

higs4281:
@saintsoup52 this PR does just what you ask -- it lets the tool go forward (status 200 means OK) despite any of those errors.

Before, we wouldn't even show the page when those errors occurred. We were just returning a nearly blank error page (with an html error status of 400) containing only the error.

You can see the old behavior right now on build. Delete the iped and refresh, and you'll get a blank page.

marteki:
@saintsoup52 to piggyback on what Bill said: if this PR doesn't send the 200 msgs, the page can't display the messaging for if there's no PID or OID, or if it's broken. @mistergone is trying to work on that, and this needs to go through first.

`$('div.verify_content').attr('data-warning')` is being populated properly when the IPED and PID and OID are missing or wonky. 👍  from me for merge.

saintsoup52:
Cool beans... Thank you for letting me know how it works :-)

mistergone:
I like it! 👍

PR #239, opened 2016-07-29
Handle non-utf-8 encoding

higs4281:
Although we specify utf-8 in our CSV spec, it is inevitable that we'll get CSVs saved straight out of Excel that are not utf-8. This should handle most such cases by decoding with latin-1 when reading in the CSV. If this doesn't work, we should reject the CSV.
## Additions
- read_in_latin_1 function
- related tests
## Testing
- With this change, the script loads the latest file from EDMC.
## Review
- @amymok 
## Checklist


amymok:
looks great

PR #240, opened 2016-07-29
Updates header to latest markup

anselmbradford:
## Changes
- Updates header markup to latest from cfgov-refresh.
## Testing
- `gulp build`
- Header should be correct at all sizes.
- There shouldn't be any header-related errors in the JS console.
## Review
- @marteki 
- @higs4281 
- @mistergone 
## Screenshots

screen shot 2016-07-29 at 6 18 40 pm
## Notes
- This has absolute URLs for the images, which doesn't seem like how it should be in production, but that's how it was before so I left it.

### DIFF
--- a/paying_for_college/templates/standalone/on-demand/header.html
+++ b/paying_for_college/templates/standalone/on-demand/header.html
@@ -3,1511 +3,1562 @@
 
 <a href="#after-nav" id="skip-nav">Skip to main content</a>
 
-<header class="o-header"
-role="banner">
-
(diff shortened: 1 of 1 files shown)


higs4281:
👍 Thank you for this, Ans.  

The absolute URLs are solely for standalone mode. In production we inherit the real header.

mistergone:
Thanks, @anselmbradford !

colbert-thumbs-up

marteki:
All the 👍 s I have. Awesome!

PR #243, opened 2016-08-01
Adjust update_ipeds to create new schools first

higs4281:
The update_ipeds command is our method of acquiring new schools that
show up in IPEDS data, and this changes the sequence of updates so that
schools are added as a first step. This allows the created schools to
benefit from the IPEDS data update that follows, so they'll have a
starter set of information that can be fleshed out later when the new
schools show up in the scorecard API.

Also fixes the help text for udpate_ipeds and its --dry-run option.
## Review
- @amymok 
## Checklist


amymok:
Looks good too!

PR #247, opened 2016-08-03
Adds option for loading program CSVs right from s3

higs4281:
Our data pipeline stores incoming school program data in s3; this allows
the load_programs script to load these files directly fro s3 rather than
requiring the moving and handling of files.
## Additions
- `--s3` option for the `load_programs` script and management command
- related tests
## Testing
- In standalone, try running load_programs with an s3 URL:

```
./manage.py load_programs --s3 true \
"http://files.consumerfinance.gov.s3.amazonaws.com/pb/paying_for_college/csv/CFPBDATAFILE713%20(2).CSV"
```

If you run `./manage.py purge programs` first, it should load 1,522 programs.  
If you run it without purging, it should report updating 1,522 programs.
## Review
- @amymok 

### DIFF
--- a/paying_for_college/management/commands/load_programs.py
+++ b/paying_for_college/management/commands/load_programs.py
@@ -1,23 +1,34 @@
 from django.core.management.base import BaseCommand, CommandError
 from paying_for_college.disclosures.scripts import load_programs
 
-COMMAND_HELP = """update_programs will update program data based on \
-data obtained from schools.  It takes in a path to a file, which \
-should be in a csv format with specific fields."""
+COMMAND_HELP = """update_programs will update program data based on
+a CSV provided by schools.  The source argument should be a CSV file path
+or, if the '--s3 true' option is passed, source should be an S3 URL."""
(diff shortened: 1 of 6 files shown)


paying_for_college/management/commands/load_programs.py
```diff
@@ -1,23 +1,34 @@
...
+        else:
+            S3 = False
+        for filesource in options['source']:
             try:
-                (FAILED, endmsg) = load_programs.load(filename)
+                if S3:
+                    (FAILED, endmsg) = load_programs.load(filesource, s3=S3)
+                else:
+                    (FAILED, endmsg) = load_programs.load(filesource)
             except:
```
amymok, line 32: To save a few lines, you can set `S3 = False` at the beginning of the function.  Then you can eliminate line 24 - 25, and you call `(FAILED, endmsg) = load_programs.load(filesource, s3=s3)` without the need of the if-else block.

Not necessary though.

amymok:
Looks good.

PR #250, opened 2016-08-03
Turn off URLs for landing page and auxiliary pages

higs4281:
This stops our incomplete pages from showing up on cfgov, but still
allows them to render in standalone mode for development.

Caveat: There are legacy assets that were stashed in
`/paying_for_college/local_static/paying_for_college` so that the site
wouldn't fall over when we first stood up standalone. Those assets are
not available in the server environments -- only standalone sees them --
so they should be tossed or moved into our front-end build
when the time comes for auxiliary-page development.
## Testing

The server part of this will have to be tested on build.  
for standalone, you shouldn't see any differences.
## Review
- @amymok @marteki @mistergone 


marteki:
The "Understanding your financial aid offer" tool is still appearing and functioning properly. The other suite pages are not, but that's to be expected right now on standalone.

👍

amymok:
👍

PR #251, opened 2016-08-04
Turn off the non-settlement disclosure page

higs4281:
In turning off URLs, I missed the bare /understanding-your-financial-aid-offer/
URL, which would probably be the starting point for nonsettlement.
## Removals
- Nonsettlement URL turned off
## Testing
- Can only be ssen on build.
- Should have no effect in standalone
## Review


marteki:
👍

PR #255, opened 2016-08-04
Scorecard updated its national stats, so we need to

higs4281:
Scorecard updated its national stats file, so we need to as well
## Changes
- Update the national_stats.json file
## Review
## Checklist


amymok:
:+1:

PR #256, opened 2016-08-05
Option to deliver generic national stats

higs4281:
In addition to delivering national salary and grad rates tied to program
length, this delivers fall-back generic salary and grad-rate values when
the bare `/api/national-stats/` endpoint is called without providing a school ID 
or program ID. This also provides a fallback in case both school ID and program ID are bad.
## Additions
- new URL option
## Changes
- StatsRepresentation view and related tests and functions
## Testing

A call to the `/api/national-stats/` [bare endpoint](http://localhost:8000/paying-for-college2/understanding-your-financial-aid-offer/api/national-stats/), with no school or program ID, should return these generic stats, including the $34,300 overall median salary used in college scorecard comparisons:

generic stats

Calls that provide an extra URL field with a schoolID_programID pair will work as before, delivering salary and grad rate values based on program length.
## Review

FYI: @saintsoup52

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ We follow the [Semantic Versioning 2.0.0](http://semver.org/) format.
 - Updates loan origination fee displays to work with corrected constants
 - Fix alignment issues in debt summary
 - Update our national statistics from scorecard project
+- Update national stats to deliver generic stats at /api/national-stats/
 
 ## 2.1.5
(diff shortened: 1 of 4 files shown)


marteki:
Values I get in browser console when running this without a school or program ID:

screen shot 2016-08-05 at 8 11 35 am

And verified that it still works for school-program pairs and solo schools.

👍

PR #274, opened 2016-08-23
Data update for tool demo

higs4281:
This PR, which is data-only (no code) introduces some technical debt by marking all current
non-settlement schools as settlement schools, in the service of enabling
demo use of the tool. The schools so tagged will carry a
"settlement_school" value of "demo," which can be identified and removed
when the false tag is no longer needed.
## Review
- Who shall give me a thumbs-up for technical debt? Show yourselves!
- @willbarton ?


willbarton:
> Who shall give me a thumbs-up for technical debt? Show yourselves!

At least we know before we owe?

:+1:

PR #275, opened 2016-08-24
Add specification docs for URLs and CSVs.

higs4281:
We need to publish our specs for data used in the tool.
This uses MkDocs to present them as gh-pages.
## Additions
- docs/ directory, spec files and MkDocs setups
## Testing
- You can run locally by installing requirements and running the MkDocs server:

```
pip install -r requirements/docs.txt
mkdocs serve
```

docs should be viewable at http://127.0.0.1:8000/
## Review
- @mistergone @marteki @amymok 
- @willbarton This is stolen straight from platform docs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,3 +74,4 @@ src/vendor/
 
 # Includes
 !src/disclosures/js/lib
+site/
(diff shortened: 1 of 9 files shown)


willbarton:
Very nice! I can't speak to the content, but everything mkdocsy looks good to me. :+1:

> @willbarton This is stolen straight from platform docs

It's not stealing if they're public domain ;)

PR #276, opened 2016-08-24
Add manage command to help tear down demo data

higs4281:
The `purge test-programs` command will wipe out fake programs created
for demo use without destroying real settlement program data and
without requiring a code deployment.
## Additions
- New parameter for the `purge` command, plus related tests
## Testing
- `./manage.py purge test-programs` could be tested against a local db, although you're unlikely to have test programs set up locally. 
  But programs are cheap and easily replaced, so there's no danger in testing. If you wipe out all programs, `./manage.py loaddata collegedata` will restore the real ones.
## Review
- @willbarton

### DIFF
--- a/paying_for_college/tests/test_scripts.py
+++ b/paying_for_college/tests/test_scripts.py
@@ -9,7 +9,7 @@ from mock import mock_open, patch
 import requests
 from django.utils import timezone
 
-from paying_for_college.models import School, Notification, Alias
+from paying_for_college.models import School, Notification, Alias, Program
 from paying_for_college.disclosures.scripts import (api_utils, update_colleges,
                                                     nat_stats, notifications,
                                                     update_ipeds,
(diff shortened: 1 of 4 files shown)


paying_for_college/tests/test_scripts.py
```diff
@@ -35,15 +35,20 @@ class PurgeTests(django.test.TestCase):
     fixtures = ['test_fixture.json', 'test_program.json']
 
     def test_purges(self):
+        self.assertFalse(Program.objects.count() == 0)
```
chosak: The Django documentation suggests using [`exists`](https://docs.djangoproject.com/en/1.8/ref/models/querysets/#exists) instead when checking for the existence of queries, so you could do something like:

``` python
self.assertFalse(Notification.objects.exists())
```

paying_for_college/tests/test_scripts.py
```diff
@@ -35,15 +35,20 @@ class PurgeTests(django.test.TestCase):
...
+        self.assertFalse(Notification.objects.count() == 0)
         self.assertTrue(purge_objects.purge('schools') ==
                         purge_objects.error_msg)
         self.assertTrue(purge_objects.purge('') ==
                         purge_objects.no_args_msg)
-        self.assertTrue("Purging" in purge_objects.purge('programs'))
+        self.assertTrue("test-programs" in
```
chosak: Alternatively could do something like:

``` python
self.assertIn('test-programs', purge_objects.purge('test-programs'))
```

paying_for_college/tests/test_scripts.py
higs4281: That's better, although in this case I think we'd need 

``` python
self.assertTrue(Notification.objects.exists())
```

We want the prove Notifications exist before wiping them.

higs4281:
@chosak thanks for the testing tips.

PR #282, opened 2016-09-28
Still some bad boxsizing.htc references surviving the build process

rosskarchner:
this uses a sed command to fix them


anselmbradford:
If it helps any, boxsizing.htc is for waaaay old IE and can be removed from the codebase altogether.

rosskarchner:
good! that dang thing is everywhere

PR #285, opened 2016-10-07
Improve message posted when updating colleges

higs4281:
This only affects a message a developer would seen when updating the
colleges database. It notes an important constant that needs to be
checked -- **Salary year** --  which is stored in the Django database. 
Salary year differs from the 'Latest year' constant, because salary data 
is always a couple years behind the rest of the data.
## Changes
- the update_colleges script, which is invoked by the `update_via_api` management command.
- the collegedata fixture includes backported updates made in production
## Review
- @amymok 
- [x] Project documentation and CHANGELOG updated.


amymok:
👍

amymok (approved):

PR #291, opened 2017-01-06
Update the `load_programs` command

higs4281:
The command's s3 loading option will now be pointed at a specific bucket, 
`validated_program_data`, to ensure that raw school submissions aren't 
inadvertently loaded into the cfgov database without being checked.

## Testing

- You should be able to run the command locally with the s3 option to load programs.  

```
python manage.py load_programs edmc.csv --s3 true
``` 

This should return "1522 programs updated."

## Review
@amymok 

* [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)


amymok:
👍

PR #296, opened 2017-01-31
Remove snyk from prepublish script

willbarton:
This PR removes snyk from the prepublish script in package.json. This should allow wheel building to happen again.

@chosak @higs4281 @ascott1


higs4281 (changes requested):
need to lose the comma at end of line 20

willbarton:
JSOOOOOOOOOOOOOOOOON!
400px-khan

higs4281 (approved):

Issue #298, opened 2017-04-23
Bug in update_colleges script


higs4281:
Two incorrect [double-equals evaluators](https://github.com/cfpb/college-costs/blob/master/paying_for_college/disclosures/scripts/update_colleges.py#L91-L94) are keeping grad rate values from populating during updates.

Addressed in

PR #299, opened 2017-04-23
Feedback properties

higs4281:
Add Feedback model properties to assist in analysis

This addresses GHE paying-for-college/dynamic-disclosures-dev issue 226

The PR also makes some minor script fixes that address #298:
- A bug in the `update_colleges` script that caused grad_rates to fail to update

## Additions
- Four model-method properties for the Feedback model
- tests for the new methods

### DIFF
--- a/paying_for_college/models.py
+++ b/paying_for_college/models.py
@@ -790,8 +790,11 @@ class Feedback(models.Model):
     def cost_error(self):
         """Return 1 or 0: Is total-cost less than tuition?"""
         url_data = self.parsed_url
-        if url_data and int(url_data['totl']) < int(url_data['tuit']):
-            return 1
+        if url_data and url_data['totl'] != '' and url_data['tuit'] != '':
+            if int(url_data['totl']) < int(url_data['tuit']):
+                return 1
+            else:
+                return 0
         else:
             return 0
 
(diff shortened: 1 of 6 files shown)


paying_for_college/models.py
```diff
@@ -739,6 +739,65 @@ class Feedback(models.Model):
...
+    @property
+    def school(self):
+        """Returns a school object, derived from a feedback url"""
+        if not self.url:
+            return ''
+        row = self.parsed_url
+        if row and row.get('iped'):
+            return School.objects.get(pk=row['iped'])
+        else:
+            return ''
```
willbarton: Is a `''` return here (and line 758) appropriate in place of a `School`? I don't see right off where this is called, so I'm not sure.

paying_for_college/models.py
```diff
@@ -739,6 +739,65 @@ class Feedback(models.Model):
...
+    def unmet_cost(self):
+        """Calculates and returns a disclosure's unmet cost"""
+        url_data = self.parsed_url
+        if not url_data:
+            return None
+
+        def total_fields(field_list):
+            total = 0
+            for field in field_list:
+                if field in url_data.keys() and url_data[field] != '':
```
willbarton: `if url_data.get('field', '') != ''` might be simpler here.

willbarton (approved):
Looks good to me overall. Nice style cleanup. Lack of familiarity with the code means I'm going over this with a fine-toothed nit picker, but neither of my comments are blockers.

higs4281:
@willbarton Thanks for the close review. suggestions taken!

PR #300, opened 2017-04-24
Fix test when using locale middleware

willbarton:
When running the college-costs tests with https://github.com/cfpb/cfgov-refresh/, the django locale middleware is loaded. That middleware chokes trying to modify the headers of the original MagicMock it got instead of a Response object. This PR removes the mocking of `render_to_response` and mocks the feedback object creation instead.

Part of GHE/CFGOV/platform/issues/798

## Changes

- Fix a test that was failing when run as part of https://github.com/cfpb/cfgov-refresh/

## Testing

- Checkout https://github.com/cfpb/cfgov-refresh/
- `pip install git+https://https://github.com/cfpb/college-costs.git@holistic-tests`
- `tox -e optional`

## Checklist

* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Passes all existing automated tests

### DIFF
--- a/paying_for_college/tests/test_views.py
+++ b/paying_for_college/tests/test_views.py
@@ -116,13 +116,12 @@ class TestViews(django.test.TestCase):
         self.assertTrue(sorted(response.context_data.keys()) ==
                         ['base_template', 'form', 'url_root'])
 
-    @mock.patch('paying_for_college.views.Feedback')
-    def test_feedback_post(self, mock_feedback):
-        response = client.post(
+    def test_feedback_post_creates_feedback(self):
+        self.assertFalse(Feedback.objects.exists())
+        client.post(
             reverse('disclosures:pfc-feedback'),
             data=self.feedback_post_data)
-        self.assertEqual(response.status_code, 200)
-        self.assertTrue(mock_feedback.call_count == 1)
+        self.assertTrue(Feedback.objects.exists())
 
     def test_feedback_post_invalid(self):
         response = client.post(reverse('disclosures:pfc-feedback'))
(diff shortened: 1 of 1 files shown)


paying_for_college/tests/test_views.py
```diff
@@ -116,12 +116,13 @@ def test_feedback(self):
...
-    def test_feedback_post(self, mock_render):
-        client.post(
+    @mock.patch('paying_for_college.views.Feedback')
+    def test_feedback_post(self, mock_feedback):
+        response = client.post(
             reverse('disclosures:pfc-feedback'),
             data=self.feedback_post_data)
-        self.assertTrue(mock_render.call_count == 1)
+        self.assertEqual(response.status_code, 200)
+        self.assertTrue(mock_feedback.call_count == 1)
```
chosak: This seems fine. One minor suggestion: this test seems to be testing that the `Feedback` constructor is called when this view is `POST`ed to. This seems a bit specific, for example the view might create the feedback object in some other way (e.g. `Feedback.objects.create(...)`). It might be more clarifying to rename this test slightly more specifically (like `test_feedback_post_creates_feedback`) and then do something like `self.assertFalse(Feedback.objects.exists())` before and then `self.assertTrue(Feedback.objects.exists())` after.

chosak (approved):

higs4281 (approved):

higs4281:
Wouldn't #299 make a handsome pairing for a release? 
??

paying_for_college/tests/test_views.py
willbarton: @chosak good point. I'm not entire sure what the original intent was here. @higs4281?

higs4281: @willbarton I join you in not being entirely sure of the author's intent, but `test_feedback_post_creates_feedback` seems clearer in this context.

willbarton: [:boom:](https://github.com/cfpb/college-costs/pull/300/commits/8c27c4e5e4ce7ef8f6b873c7ed80d45662ace109)

richaagarwal (approved):

PR #307, opened 2017-06-01
Catch missing URL values in cost-error model method

higs4281:
We've encountered a few disclosure URLs that were missing some fields.
This uses the dictionary get() method, which should have been used in
the first place, to avoid key errors when running reports.

### DIFF
--- a/paying_for_college/models.py
+++ b/paying_for_college/models.py
@@ -400,8 +400,10 @@ class DisclosureBase(models.Model):
     def cost_error(self):
         """Return 1 or 0: Is total-cost less than tuition?"""
         url_data = self.parsed_url
-        if url_data and url_data['totl'] != '' and url_data['tuit'] != '':
-            if int(url_data['totl']) < int(url_data['tuit']):
+        if (url_data
+                and url_data.get('totl') != ''
+                and url_data.get('tuit') != ''):
+            if int(url_data.get('totl', 0)) < int(url_data.get('tuit', 0)):
                 return 1
             else:
                 return 0
--- a/paying_for_college/models.py
+++ b/paying_for_college/models.py
@@ -400,10 +400,10 @@ class DisclosureBase(models.Model):
     def cost_error(self):
         """Return 1 or 0: Is total-cost less than tuition?"""
         url_data = self.parsed_url
-        if (url_data
-                and url_data.get('totl') != ''
-                and url_data.get('tuit') != ''):
-            if int(url_data.get('totl', 0)) < int(url_data.get('tuit', 0)):
+        if url_data:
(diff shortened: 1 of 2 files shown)


paying_for_college/models.py
```diff
@@ -400,8 +400,10 @@ def total_fields(field_list):
     def cost_error(self):
         """Return 1 or 0: Is total-cost less than tuition?"""
         url_data = self.parsed_url
-        if url_data and url_data['totl'] != '' and url_data['tuit'] != '':
-            if int(url_data['totl']) < int(url_data['tuit']):
+        if (url_data
+                and url_data.get('totl') != ''
+                and url_data.get('tuit') != ''):
+            if int(url_data.get('totl', 0)) < int(url_data.get('tuit', 0)):
```
richaagarwal: Minor thing, but since you call `url_data.get('totl')` and `url_dat.get(tuit')` more than once, could you store them in a variable?

richaagarwal (approved):
Left one comment but otherwise looks good!

paying_for_college/models.py
richaagarwal: Another thing is, instead of doing `if ... return 1` else `return 0`, you could do `return int(total < tuition)` but your call on if that's more readable or not

higs4281:
@richaagarwal Thanks! Your suggestions led to some better edge-case tests.

paying_for_college/models.py
```diff
@@ -400,8 +400,10 @@ def total_fields(field_list):
     def cost_error(self):
         """Return 1 or 0: Is total-cost less than tuition?"""
         url_data = self.parsed_url
-        if url_data and url_data['totl'] != '' and url_data['tuit'] != '':
-            if int(url_data['totl']) < int(url_data['tuit']):
+        if url_data:
+            totl = url_data.get('totl') or 0
+            tuit = url_data.get('tuit') or 0
```
richaagarwal, line 405: Oh yeah, good call on assigning them a value, since converting `None` to `int` would've inevitably led to an error. Sorry I missed that in my review. 

Not to be too nitpicky, but I think it's more canonical to do `url_data.get('totl', 0)` to assign a default value.

higs4281: Not nitpicky. I started with `url_data.get('totl', 0)` but that would allow an empty string to pass through, and I actually want to coerce both empty string and None to zero.

richaagarwal: Oh that makes sense, got it 👍🏽

willbarton: Popping in to say I wouldn't have caught that either! Clever.

PR #308, opened 2017-08-23
Bug Fixes

mistergone:
This pull request fixes four bugs in the tool:
- Definition lists aren't stacking properly (see GHE platform issue 1054)
- Browser forcing auto-scroll down in small screen (see GHE platform issue 1055)
- Tab navigation within budgeting worksheet wipes out edits(see GHE platform issue 1056)
-  Spacing misalignment for a <p> within a <div> (see GHE platform issue 1057)

## Changes
- Added some LESS so that `dd` and `dt` stack as originally intended in the verification section
- Disabled the `sticky-kit.js` events when the screen is smaller than 600px wide. This fixes the chaotic auto-scrolling issue.
- Rewrote the `keyup` handler for the Expenses section so that it matches the code for the financial section. This fixes the issue with tabbing wiping out edits.
- Added LESS to fix the `margin-bottom` on the `p:last-child` elements. (This was caused by cascading styles bleeding in from `header.css` and `footer.css`.)

## Testing
- Pull down and install `college-costs` as directed in the README. Then look at the internal issues listed above and check my fixes!
- __IMPORTANT NOTE__: `college-costs` does __NOT__ work with node 8, and `gulp` commands will not work under node 8. Since many of our other apps now use node 8, be sure you are using node 6. (I have not tested in any other node versions).

## Review
- @sebworks has promised to check my code
- @marteki - if possible, I would like for you to review the work and give it a thumbs-up.

## Todos
- The app should be updated to work in node 8 at some point. This might be as simple as updating `gulp`, but I didn't look into it.

## Checklist
* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Passes all existing automated tests


marteki (approved):
I reviewed the end result after running it locally, and not the code. All bugs fixed. :+1: from me.

Totally out of scope for this PR: adding a `.nvmrc` to this project, which I think would make sure anyone working in the code for it would automatically use Node 6 rather than Node 8 (until it's updated to use 8).

sebworks:
I don't think the shrinkwrap should change.

src/disclosures/js/views/financial-view.js
```diff
@@ -804,14 +804,23 @@ var financialView = {
...
       } )
       .on( 'sticky_kit:unbottom', function( evt ) {
         $( evt.target ).removeClass( 'is_bottomed' );
       } );
+    } else {
+      $stickyOffers.trigger( 'sticky_kit:detach' );
+    }
+
+    $win.resize( function() {
+      financialView.stickySummariesListener();
```
sebworks: If you add the function call in the resize function, the the event listener is going to added many times.

mistergone: Oops, you're right. That's an order of operations error on my part. I'll move the `detach` trigger so that it always happens first when the function is called.

I'll discuss it with you later this morning to make sure it makes sense.

mistergone:
@sebworks approved my additional changes, so I am now merging!

PR #315, opened 2018-04-02
Use hyphenated name for cfgov-setup import

higs4281:
This uses the standard name as specified in the [cfgov-django-setup repo](https://github.com/cfpb/cfgov-django-setup#installation) for the setup.py config.

This does not need to be deployed and can wait for the next substantive change for a release.
The current config will work thanks to permissive pypi name-handling, but we should use the proper name for our own package.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,6 +56,6 @@ setup(
     ],
     long_description=read_file('README.md'),
     zip_safe=False,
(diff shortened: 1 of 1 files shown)


richaagarwal (approved):

chosak (approved):
LGTM. FWIW, this [is not the only repo](https://github.com/search?q=user%3Acfpb+cfgov_setup&type=Code) that is guilty of mishyphenation.

PR #322, opened 2018-07-26
Modernizes build

anselmbradford:
Someone with familiarity with college-costs should try this out. I'm not sure exactly what to check.

## Changes

- Convert browserify config to webpack.
- Remove unused dependencies.
- Update ESLint config to version 5.
- Fix linter errors.
- Convert to ES6 syntax.
- Add linting to Travis config.

## Testing

- https://github.com/cfpb/college-costs#installation should work (specifically `./standalone_setup.sh`).
- `npm test` should pass.
- `gulp lint` should not have errors (it will have warnings).
- The protractor tests in https://github.com/cfpb/college-costs#running-tests don't pass for me, but it looks like maybe I need an API running. Anyone able to check?


mistergone (approved):
👍

higs4281 (approved):
I ran into a problem with a migration that blocked setup for me. I opened a PR to your PR that tweaks the migration to locally sidestep an irrelevant error and allow setup to proceed. It shouldn't affect anything else. I tested this change inside cfgov and no new migrations run.

With migrations fixed, I had to load the base data to get protractor tests to run:

```
```


After that, setup worked, the base page rendered and I was able to build a deployment wheel, which runs a front-end build.

All tests ran, including the browser tests, although many of those  failed. I think the browser tests may be making  assumptions about test data that setup isn't providing. This is a browser testing issue, not an issue with your build PR.

PR #323, opened 2018-08-01
Fix migration for stand-alone usage.

higs4281:
This just allows local data migrations to sidestep an error that can crop up
while setting up a standalone version. It shouldn't affect any other operations.


anselmbradford:
How do I test? `./standalone_setup.sh`?

anselmbradford:
Also, @mistergone is in here right now testing my PR, so might be easy for him to try.

higs4281:
You can test the migration by running migrate:

```
./manage.py migrate
```

If that succeeds, then loading data should work as well.

```
```
(loaddata takes a minute)

higs4281:
If you've recently worked on cfgov-refresh, you may need to unset DJANGO_SETTINGS_MODULE, which might get in your way.

```
unset DJANGO_SETTINGS_MODULE
```

anselmbradford (approved):

Issue #325, opened 2018-08-15
Snyk vulnerabilities


chosak:
The `npm test` command fails because its subcommand `snyk test` fails with several high severity vulnerabilities:

> Tested 586 dependencies for known vulnerabilities, found 20 vulnerabilities, 69 vulnerable paths.

To reproduce, first log in to [Snyk](https://snyk.io/) using `snyk auth` and then run `snyk test` from the command line. Alternatively, you can browse the vulnerabilities through the [Snyk web UI](https://app.snyk.io/org/cfpb) (requires membership in the "cfpb" organization).

The high severity vulnerabilities seem to come from `student-debt-calc@2.6.3`, which is also a CFPB package from https://github.com/cfpb/student-debt-calculator. Correcting the vulnerabilities here may require a new version release of that package that corrects its vulnerabilities.

chosak:
Now that @mistergone has merged #328, some of these vulnerabilities have been fixed, but one still remains when running `npm test`:

```
✗ Medium severity vulnerability found in jquery
  Description: Cross-site Scripting (XSS)
  Info: https://snyk.io/vuln/npm:jquery:20150627
  Introduced through: jquery@1.11.3, jquery.easing@1.3.2
  From: jquery@1.11.3
  From: jquery.easing@1.3.2 > jquery@1.11.3
  Remediation:
    Your dependencies are out of date, otherwise you would be using a newer version of jquery. 
    Try deleting node_modules, reinstalling and running `snyk test` again. If the problem persists, one of your dependencies may be bundling outdated modules.

Organisation:    chosak
Package manager: npm
Target file:     package.json
Open source:     no
Project path:    /Projects/college-costs
Local Snyk policy: found

Tested 327 dependencies for known vulnerabilities, found 1 vulnerability, 2 vulnerable paths.

Run `snyk wizard` to address these issues.

npm ERR! Test failed.  See above for more details.
```

One option would be to update this repo's [.snyk file](https://github.com/cfpb/college-costs/blob/master/.snyk) to ignore this jQuery vulnerability. We could also change the `npm test` command so that it runs `snyk test --severity-threshold=high`.

chosak:
Closed by #332:

```sh
$ npm run snyk

> college-costs@ snyk /Users/chosaka/Projects/college-costs
> snyk test


Testing /Users/chosaka/Projects/college-costs...

Organisation:    chosak
Package manager: npm
Target file:     package.json
Open source:     no
[...]

PR #328, opened 2018-08-17
Update dependencies

mistergone:
- Update front-end dependencies that are marked as vulnerabilities

## Changes
- Changes to `package.json` based on vulnerability reports from `snyk` and `npm audit`
- Change `capital-framework` dependency to dependencies to the `cf-*`
components individually
- Update webfont calls in LESS to use variables

## Review
- @chosak


src/disclosures/css/cf-enhancements.less
```diff
@@ -319,7 +319,7 @@ a.btn__full-small.btn__link {
                  unit( 6px / @base-font-size-px, em );
         border: 0;
         outline: 2px solid transparent;
-        .webfont-regular();
```
anselmbradford, line 322: This was updated to `.u-webfont-regular()`. However, it has since been [deprecated](https://github.com/cfpb/capital-framework/blob/master/src/cf-core/usage.md#webfont-mixins). You may still want to use it though in place of setting the font value directly, so that it's easier to find and strip out in the future.

schbetsy (approved):

chosak:
@mistergone this works as advertised, however: when running `npm test` (which runs `snyk test`) on this branch:

```
Testing /Users/chosaka/Projects/college-costs...

✗ Medium severity vulnerability found in jquery
  Description: Cross-site Scripting (XSS)
  Info: https://snyk.io/vuln/npm:jquery:20150627
  Introduced through: jquery@1.11.3, jquery.easing@1.3.2
  From: jquery@1.11.3
  From: jquery.easing@1.3.2 > jquery@1.11.3
  Remediation:
    Your dependencies are out of date, otherwise you would be using a newer version of jquery. 
    Try deleting node_modules, reinstalling and running `snyk test` again. If the problem persists, one of your dependencies may be bundling outdated modules.

Organisation:    chosak
Package manager: npm
Target file:     package.json
Open source:     no
Project path:    /Projects/college-costs
Local Snyk policy: found

Tested 327 dependencies for known vulnerabilities, found 1 vulnerability, 2 vulnerable paths.

Run `snyk wizard` to address these issues.

npm ERR! Test failed.  See above for more details.
```

Note that this doesn't fail Travis because Travis doesn't run `npm test`.

One option would be to update this repo's [.snyk file](https://github.com/cfpb/college-costs/blob/master/.snyk) to ignore this jQuery vulnerability. We could also change the `npm test` command so that it runs `snyk test --severity-threshold=high`. Either of these things could be done as part of this PR or as a subsequent PR -- deferring to @mistergone on how best to handle.

mistergone:
How do you feel about a subsequent PR where I check into updating jQuery?

chosak:
:+1: Why don't you merge this one and then I'll update #325 with my comment above, and we can keep that one open to track that issue.

chosak:
@mistergone another thing I just noticed: after running `./setup.sh` I get a modified `package-lock.json` file. If the frontend setup causes changes to the local files it will complicate the wheel names on Travis unless we do a fix similar to [teachers-digital-platform#213](https://github.com/cfpb/teachers-digital-platform/pull/213).

src/disclosures/css/cf-enhancements.less
mistergone: I updated these to variables - thanks for the note!

anselmbradford: Cool, thanks!

PR #333, opened 2018-09-11
Update Library Versions

schbetsy:
Ensure that the set of allowed versions contains the actual version that cfgov-refresh is currently on, so when we import and install this repo in cfgov, it doesn't cause version conflicts.

For most libraries, allow any revision within a major version. There are a few exceptions, especially for libraries that drop compatibility with django 1.8 without a major version bump.

## Testing

`tox -r` will run the tests on Django 1.8 and 1.11 with upgraded versions of all affected libraries.

## Checklist

* [x] Changes are limited to a single goal (no scope creep)
* [x] Code can be automatically merged (no conflicts)
* [x] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [x] Passes all existing automated tests
* [x] New functions include new tests
* [x] New functions are documented (with a description, list of inputs, and expected output)
* [x] Placeholder code is flagged
* [x] Visually tested in supported browsers and devices
* [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,29 @@ apache-jmeter-3.0/bin/jmeter.sh -t college-costs/test/load_testing/Offer.jmx -Js
 
 The corresponding `Offer.csv` in `test/load_testing` specified the parameters being used for testing.  The order and name of each column is specified in the JMX file under CSV Offer.
 
+## API Docs
+
+[Documentation](https://cfpb.github.io/college-costs/) for this repository is rendered via GitHub pages and [Swagger](https://swagger.io/docs/). They can be edited in the `docs/` directory, but to view or deploy them, you'll need to install the docs dependencies listed in the `setup.py`:
+
+```
+pip install -e '.[docs]'
+```
+
+You can then preview your changes locally by running `mkdocs serve` and then reviewing <http://127.0.0.1:8000/>
+
+When your changes are ready, you can submit them as a normal pull request. After that, you can use this command to publish them:
+
+```
+mkdocs gh-deploy --clean
+```
+
+That pushes the necessary files to the `gh-pages` branch.
+
+### Notes
+
+- The `mkdocs gh-deploy` command will push any edits you've made locally to the `gh-pages` branch of the repository, whether you've committed them or not.
+- Mkdocs will create a "site" directory locally when you use either `build`, `serve` or `gh-deploy`. This is used to assemble assets to be pushed to the gh-pages branch, but the `site/` directory itself doesn't need to be under version control. It can be deleted after a deploy.
+
 ## Getting involved
 
 If you find a bug or see a way to improve the project, we'd love to hear from you. Add an issue, or fork the project and send us a pull request with your suggested changes.
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,6 @@ install_requires = [
     'djangorestframework==3.6.4',
     'elasticsearch>=2.4.1,<3',
     'PyYAML>=3.11,<3.14',
-    'python-dateutil>=2.2,<3',
     'requests>=2.18,<3',
     'Unipath>=1.1,<2',
 ]
--- a/setup.py
+++ b/setup.py
@@ -10,26 +10,26 @@ setup_requires=[
 
 install_requires = [
     'Django>=1.8,<1.12',
-    'django-haystack==2.7.0',
+    'django-haystack==2.7.0',  # Latest version that supports both Django 1.8 and 1.11
     'djangorestframework==3.6.4',
-    'elasticsearch==2.4.1',
-    'PyYAML==3.11',
-    'python-dateutil==2.2',
-    'requests==2.7.0',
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -12,24 +12,24 @@
...
+    'PyYAML>=3.11,<3.14',
+    'python-dateutil>=2.2,<3',
+    'requests>=2.18,<3',
+    'Unipath>=1.1,<2',
 ]
 
 
 docs_extras = [
-    'Markdown==2.3.1',
+    'Markdown==2.6.9',
```
schbetsy: For the docs requirements, I've only upgraded them to match cfgov-refresh instead of doing the whole process of researching each library to find its versioning scheme so I can decide what version range to list. I don't feel very confident making more changes to docs requirements, since they're not covered by tests. 

Another option would be to leave them alone entirely (and undo the changes I've made here), since docs requirements aren't pulled in to cfgov-refresh, so version mismatches won't cause conflicts. Do you have opinions on what I should do here, @chosak ?

chosak: I don't feel strongly, although I notice two things that are strange:

1. Running `pip install -e '.[docs,testing]'`, I see these warnings:

   > requests 2.19.1 has requirement certifi>=2017.4.17, but you'll have certifi 2016.8.2 which is incompatible.
mkdocs 0.16.3 has requirement livereload>=2.5.1, but you'll have livereload 2.3.2 which is incompatible.

2. The PyYAML versions in `install_requires` and `docs_extras` seem to conflict, although they also do [in cfgov-refresh](https://github.com/cfpb/cfgov-refresh/blob/master/requirements/manual.txt#L2).

If you look in [the `setup.py` requirements for mkdocs](https://github.com/mkdocs/mkdocs/blob/0.16.3/setup.py#L60), the versions there are also different from the ones here in `docs_extras`. I wonder where the list of docs requirements came from here, maybe someone did a `pip freeze` on their virtualenv and just stuck those versions in? If I try with this much-reduced list of requirements, keeping only mkdocs and the mkDOCter template:

```py
docs_extras = [
    'mkDOCter==1.0.5',
    'mkdocs==0.16.3',
]
```

then commands like `mkdocs build` and `mkdocs serve` still work to build and serve all of the documentation, seemingly correctly. So maybe we can just reduce these down.

Then again, I agree with you that it's not great that (a) these aren't covered by tests and (b) isn't mentioned in the README, neither how to build locally or how to push changes to the GH-hosted version. So, while we could just leave as-is, it'd be nice to get this working/documented, while we're in here, if it's not too much work.

schbetsy: Excellent idea. Done in a0b4556 and 3bf0504.

README.md
```diff
@@ -136,6 +136,29 @@ apache-jmeter-3.0/bin/jmeter.sh -t college-costs/test/load_testing/Offer.jmx -Js
...
+
+```
+mkdocs gh-deploy --clean
+```
+
+That pushes the necessary files to the `gh-pages` branch.
+
+### Notes
+
+- The `mkdocs gh-deploy` command will push any edits you've made locally to the `gh-pages` branch of the repository, whether you've committed them or not.
```
chosak, line 159: Thanks for the thorough documentation on this!

setup.py
```diff
@@ -2,45 +2,33 @@
...
+    'django-haystack==2.7.0',  # Latest version that supports both Django 1.8 and 1.11
     'djangorestframework==3.6.4',
-    'elasticsearch==2.4.1',
-    'PyYAML==3.11',
-    'python-dateutil==2.2',
-    'requests==2.7.0',
-    'Unipath==1.1',
+    'elasticsearch>=2.4.1,<3',
+    'PyYAML>=3.11,<3.14',
+    'python-dateutil>=2.2,<3',
```
chosak: Super minor and annoying thing: in cfgov-refresh this is currently [`==2.1`](https://github.com/cfpb/cfgov-refresh/blob/138b03b5e400aa539409cd7b0700cf71390a7253/requirements/libraries.txt#L42). The changelog entry for 2.2 is [here](https://dateutil.readthedocs.io/en/stable/changelog.html#version-2-2). I don't really know enough about that library at first glance to know if those are big things we need to worry about or not. But:

It looks to me like the only use of `dateutil` in this whole repository is in [this unit test file](https://github.com/cfpb/college-costs/blob/a75769f0df9d1ceabaf991cf83e4dba8b48c308b/paying_for_college/tests/test_models.py#L351). So perhaps this could/should be moved out to `testing_extras`? If you look into that test file, it looks to me like that library could even be removed in favor of just creating the timestamps directly. So maybe we could even just strip it out?

schbetsy: Well spotted. Does this work: f340146? (`tox` passes, so I hope so...)

chosak: I see, so the tests don't even use that timestamp at all. Nice fix.

chosak (approved):

PR #335, opened 2018-09-19
Include cf-icons SVGs as static assets in build

chosak:
The use of Capital Framework cf-forms in this project requires inclusion of the cf-icons icon SVG files. This change adds a gulp copy step to copy these to the right static output folder and also includes them in the Python wheel build.

See https://github.com/cfpb/capital-framework/issues/864.


mistergone (approved):
Looks good to me!

PR #342, opened 2019-03-27
Update pyyaml to 3.13 to match cfgov-refresh

higs4281:
Nail version down.


Scotchester:
Snyk can be ignored, since it's testing this branch that is a little behind master, where we've updated the dependencies it's watching.

Scotchester (approved):

PR #343, opened 2019-04-03
Py3 testing

higs4281:
Code refactor for Python 3 and linting compliance.
~~I'm still fiddling with some csv tests and the coverage drop, but t~~ 
The beast can be observed.

[edited: imports combined]
~~It occurred to me that I do a variation of this little dance in a number of places, and it would be nice to just do it once:~~
~~import six~~
~~if six.PY2:~~
    ~~from unicodecsv import DictReader~~
~~else:~~
    ~~from csv import DictReader~~


paying_for_college/disclosures/scripts/nat_stats.py
```diff
@@ -41,7 +42,7 @@ def get_stats_yaml():
         nat_yaml = requests.get(COLLEGE_CHOICE_NATIONAL_DATA_URL)
         if nat_yaml.ok and nat_yaml.text:
             nat_dict = yaml.safe_load(nat_yaml.text)
-    except:
+    except Exception:
```
willbarton: Do we know what exceptions might be raised to be excepted here? Just ones from `requests.get()` and `yaml.safe_load()`?

paying_for_college/models.py
```diff
@@ -89,8 +95,8 @@ class ConstantRate(models.Model):
     note = models.TextField(blank=True)
     updated = models.DateField(auto_now=True)
 
-    def __unicode__(self):
-        return u"%s (%s), updated %s" % (self.name, self.slug, self.updated)
+    def __str__(self):
```
willbarton, line 100: Should this class get a corresponding [`@python_2_unicode_compatible`](https://docs.djangoproject.com/en/1.10/topics/python3/#str-and-unicode-methods)? i.e.

```python
from django.utils.encoding import python_2_unicode_compatible

@python_2_unicode_compatible
class ConstantRate(models.Model):
```

paying_for_college/models.py
```diff
@@ -99,15 +105,16 @@ class Meta:
...
+        max_length=255,
+        blank=True,
+        help_text="VARIABLE NAME FOR JS")
     value = models.IntegerField()
     note = models.TextField(blank=True)
     updated = models.DateField(auto_now=True)
 
-    def __unicode__(self):
-        return u"%s (%s), updated %s" % (self.name, self.slug, self.updated)
+    def __str__(self):
```
willbarton, line 119: Same as above

paying_for_college/models.py
```diff
@@ -169,9 +176,10 @@ class Contact(models.Model):
     name = models.CharField(max_length=255, blank=True)
     internal_note = models.TextField(blank=True)
 
-    def __unicode__(self):
-        return u", ".join([bit for bit in [self.contacts,
-                                           self.endpoint] if bit])
+    def __str__(self):
```
willbarton, line 183: Same as above

paying_for_college/models.py
```diff
@@ -292,28 +305,28 @@ def as_json(self):
             ordered_out[key] = dict_out[key]
         return json.dumps(ordered_out)
 
-    def __unicode__(self):
-        return self.primary_alias + u" (%s)" % self.school_id
+    def __str__(self):
```
willbarton, line 313: Same with respect to `@python_2_unicode_compatible`

paying_for_college/models.py
```diff
@@ -440,29 +453,32 @@ class Notification(DisclosureBase):
     sent = models.BooleanField(default=False)
     log = models.TextField(blank=True)
 
-    def __unicode__(self):
-        return "{0} {1} ({2})".format(self.oid,
-                                      self.institution.primary_alias,
-                                      self.institution.pk)
+    def __str__(self):
```
willbarton, line 462: Same as above

paying_for_college/models.py
```diff
@@ -539,8 +558,8 @@ class Disclosure(models.Model):
     institution = models.ForeignKey(School, blank=True, null=True)
     text = models.TextField(blank=True)
 
-    def __unicode__(self):
-        return self.name + u" (%s)" % unicode(self.institution)
+    def __str__(self):
```
willbarton, line 568: Same as above

paying_for_college/models.py
```diff
@@ -611,8 +630,8 @@ class Program(models.Model):
                                 help_text="EXPLANATION FROM SCHOOL")
     test = models.BooleanField(default=False)
 
-    def __unicode__(self):
-        return u"%s (%s)" % (self.program_name, unicode(self.institution))
+    def __str__(self):
```
willbarton, line 641: Same as above

paying_for_college/models.py
```diff
@@ -776,8 +795,8 @@ class Alias(models.Model):
     alias = models.TextField()
     is_primary = models.BooleanField(default=False)
 
-    def __unicode__(self):
-        return u"%s (alias for %s)" % (self.alias, unicode(self.institution))
+    def __str__(self):
```
willbarton, line 807: Same as above

paying_for_college/models.py
```diff
@@ -791,9 +810,9 @@ class Nickname(models.Model):
     nickname = models.TextField()
     is_female = models.BooleanField(default=False)
 
-    def __unicode__(self):
-        return u"%s (nickname for %s)" % (self.nickname,
-                                          unicode(self.institution))
+    def __str__(self):
```
willbarton, line 823: Same as above

paying_for_college/tests/test_commands.py
```diff
@@ -1,33 +1,40 @@
...
         stdout_patch.start()
         self.addCleanup(stdout_patch.stop)
 
-    @mock.patch('paying_for_college.management.commands.'
-                'update_pfc_national_stats.nat_stats.'
-                'update_national_stats_file')
+    @mock.patch(
+        'paying_for_college.management.commands.'
+        'update_pfc_national_stats.nat_stats.'
+        'update_national_stats_file')
```
willbarton, line 23: Maybe something for broader BEWD discussion, but I feel like I'm okay with keeping long paths like these whole and using `# noqa`. Doesn't mean you have to change these.

paying_for_college/views.py
```diff
@@ -1,32 +1,36 @@
...
+from paying_for_college.models import (
+    ConstantCap, ConstantRate, Feedback, Notification, Program, School,
+    Worksheet
+)
+
+
+try:
+    from collections import OrderedDict
+except Exception:  # pragma: no cover
+    from ordereddict import OrderedDict
```
willbarton: Looks like you removed the other `ordereddict.OrderedDict` imports. (`collections.OrderedDict` is in 2.7.)

setup.py
```diff
@@ -78,8 +79,8 @@ def read_file(filename):
         'Topic :: Internet :: WWW/HTTP',
         'Intended Audience :: Developers',
         'Programming Language :: Python',
-        'Programming Language :: Python :: 2.6',
         'Programming Language :: Python :: 2.7',
+        'Programming Language :: Python :: 3.6',
```
willbarton, line 83: This change makes me laugh.

willbarton (approved):
Whew. Finally made a read-through. The changes all make sense to me and the tests pass for me! Heroic effort! A few comments follow.

paying_for_college/disclosures/scripts/nat_stats.py
higs4281: We should watch for yaml AttributeError (if request.text is None) and requests ConnectionError if the scorecard URL changes. Added explicit exceptions and tests.

paying_for_college/tests/test_commands.py
higs4281: I felt that way at first, but then, for patches with long paths, I found it easier to track.  
I'd be happy either way.

paying_for_college/views.py
higs4281: Let us lose this 2.6 relic!

higs4281:
@willbarton Thank you for the review, and for catching the python_2_unicode_compatible omissions!

PR #349, opened 2019-05-14
Pass the offer data explicitly to the feedback form

willbarton:
The feedback form had previously acquired the offer data indirectly from the referrer URL. This change passes that data directly to the form in the form of the query string so that it can be recorded along with the message. This should fix the current issues of the referrer being blank by explicitly providing the data needed.

## Checklist

* [x] Changes are limited to a single goal (no scope creep)
* [x] Code can be automatically merged (no conflicts)
* [x] Passes all existing automated tests


schbetsy (approved):
With this change, I was able to submit feedback through the form, and I saw that the query string (including program information) was saved in the `url` field of the feedback. LGTM.

PR #350, opened 2019-05-17
Stop the Travis from giving our wheels a dirty name

higs4281:
Package-lock changes during Travis builds can result in wheels with polluted names, such as:

college_costs-2.6.1.dev0+3f1ded0-py2.py3-none-any.whl

Since this repo is likely to move soon into cfgov-refresh, with its yarn build setup, it makes sense to me to let Travis sidestep package-lock.json for the time being.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,11 @@ python:
 install:
 - pip install tox-travis
 - npm install -g gulp-cli
-- npm config set package-lock false
 - pip install coveralls tox
 script:
+- npm config set package-lock false
 - tox
 - python2.7 setup.py bdist_wheel --universal
-- npm config set package-lock false
 - npm test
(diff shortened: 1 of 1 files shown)


chosak (commented):
Strangely, this line is already in this repo [here](https://github.com/cfpb/college-costs/pull/350/files#diff-354f30a63fb0907d4ad57269548329e3R13)! Does it just need to be moved up two lines, before the call to tox?

higs4281:
> Strangely, this line is already in this repo [here](https://github.com/cfpb/college-costs/pull/350/files#diff-354f30a63fb0907d4ad57269548329e3R13)! Does it just need to be moved up two lines, before the call to tox?

Let's try that.

chosak (approved):

PR #351, opened 2019-05-31
Fix the capturing of school closures.

higs4281:
We weren't saving the College Scorecard API's `school.operating` value properly,
because the delivered values are 1 and 0, not True or False.


mistergone (approved):
👍

PR #356, opened 2019-06-21
Set icon paths | update cf-select to a-select

anselmbradford:
https://github.com/cfpb/college-costs/pull/355 updated cf-forms to 10.2.0, which utilizes the erroneously unused `@cf-icon-path` variable, which subsequently had the wrong path set in this repo.

## Removals

- Deprecated `cf-select` styles.
- Deprecated `@cf-icon-prefix` and `@cf-icon-ie7-support` variables.

## Changes

- Update `@cf-icon-path` variable to correct path.

## Testing

- Run local setup or in cfgov-refresh and http://localhost:8000/paying-for-college2/understanding-your-financial-aid-offer/offer/ select drop-down menu should work as before.


chosak (approved):
Works for me. I can confirm that running collectstatic locally (with production settings) in cfgov-refresh works on this branch, but not on the current master here. So that _should_ mean it'll work for real.

PR #357, opened 2019-06-26
Add deprecation note

higs4281:
The college-costs code now lives as part of the cfgov-refresh repo, but
we need to maintain, at least for now, the published documentation for schools
that use the settlement version of the tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,9 @@
 # Paying for College
 Tools to help students make informed financial decisions about college.
 
+## ⚠️ DEPRECATION NOTE ⚠️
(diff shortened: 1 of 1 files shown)


Scotchester:
If we need to maintain this repo for docs (and have it be unarchived if we need to make edits), should we just delete the master branch and make the gh-pages branch the main branch?

higs4281:
> If we need to maintain this repo for docs (and have it be unarchived if we need to make edits), should we just delete the master branch and make the gh-pages branch the main branch?

I think I'd vote for archiving as-is, to leave the code as a read-only record of a public tool built under unusual circumstances. We'll know soon whether the docs need any changes.

